PR #151, opened 2013-11-14
Multistep measurement

iteratec-hh:
Extension for webpagetest to deliver and show multiple results from a single (multistep-)measurement. So the test-frequency can be increased, clearly with fix agent-capacity.

The single results are displayed in web-gui of webpagetest. We adapted all subpages of web-gui to show the multistep-results. We developed the solution to fit the needs of one of our customers. We believe, the ability to measure multiple steps within a single test could be very useful for other wpt-users, too.

We adapted wptdriver, Browser-extensions and PHP-code of the web-gui. We didn't adapt urlblast.

--
iteratec GmbH (http://www.iteratec.de)


pmeenan:
This has been updated in #376 - closing the old pull request

PR #375, opened 2015-01-06
updated documentation for auto scaling and instance configuration

pherris:
These suggested edits are based on my experience setting up the AMI and will hopefully save a few forum posts. 

Note, that I am only assuming the docs were wrong when they say the server code is updated hourly.  I made some changes manually on the server and those changes have not been overwritten in 24+ hrs.


pmeenan:
The server updates use "git pull origin master" to use git to sync with the head on github.  They won't overwrite any local changes (they will usually merge).  You'll need to "git reset --hard HEAD" or something like that to throw away the modifications you made on the server if you want to go back to vanilla.

pmeenan:
Mostly looks good if you could just put the bit about the server updating back in.

pherris:
Thanks, PR updated to include the server automatic update.

PR #381, opened 2015-01-14
Added JSON option to domains.php

andydavies:
Let me know if you want to approach it another way and I'll refactor the code.

(not sure why the GH diff is marking things that haven't changed as being changed)


pmeenan:
Mostly good (I think the big diff is because of line ending differences).

Only 2 changes I'd like to see:

1 - Keep domains.inc as a require_once in case common.inc ends up including it at some point.
2 - Change the "return" at the end of the json processing to an exit

andydavies:
Updated…

PR #384, opened 2015-01-18
added tag logic

flyinprogrammer:
This patch allows you to inject configurable static tags when test agent instances are created. This might be helpful for people who are trying to use tags to identity what instances are being used for what purposes.

Starting the AMI, or adding a setting like this to your settings.ini, will result in these tags being added to the test agent instances that are launched:

EC2.tags="department=>performance|project=>new_featureX,new-featureY"

I chose a hash rocket, pipe syntax because sometimes you want to use commas and hyphens in your tag values, and building a complex parser seemed unnecessary.

### DIFF
--- a/www/ec2/ec2.inc.php
+++ b/www/ec2/ec2.inc.php
@@ -637,7 +637,7 @@ function EC2_CreateTagArray($tagstring) {
 
   foreach ($kvpairs as $kvpair) {
     $pair = explode('=>', $kvpair);
-    array_push($final_array, array('Key' => $pair[0], 'Value' => $pair[1]));
+    $final_array[] = array('Key' => $pair[0], 'Value' => $pair[1]);
   }
   return $final_array;
 }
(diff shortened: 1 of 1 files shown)


www/ec2/ec2.inc.php
```diff
@@ -593,11 +597,48 @@ function EC2Log($msg) {
...
+ * @param string $tagstring A tag delimited string.
+ * @return array An array of array tag key-value pairs
+ */
+function EC2_CreateTagArray($tagstring) {
+  $kvpairs = explode('|', $tagstring);
+  $final_array = array();
+
+  foreach ($kvpairs as $kvpair) {
+    $pair = explode('=>', $kvpair);
+    array_push($final_array, array('Key' => $pair[0], 'Value' => $pair[1]));
```
pmeenan: nit: could you make this:
$final_array[] = array('Key' => $pair[0], 'Value' => $pair[1]);

Everything else looks great.

flyinprogrammer:
anything i can do to make this code better?

pmeenan:
Sorry, missed the update.  LGTM.

PR #392, opened 2015-01-29
allow clearing out disk space without archiving

ovesh:
Specifying archive_dir='none' in settings.ini will now skip archiving
in www/cli/archive.php

See discussion in http://www.webpagetest.org/forums/showthread.php?tid=13516&pid=25469

### DIFF
--- a/www/archive.inc
+++ b/www/archive.inc
@@ -17,6 +17,9 @@ function ArchiveTest($id, $store_result = true, $delete = false) {
     if ( strpos($id, '.') === false &&                           // make sure it isn't a relay test
         (array_key_exists('archive_dir', $settings) ||           // local mount
         array_key_exists('archive_s3_server', $settings)) ) {    // S3 Interface (Internet Archive or standard)
+        if ($settings['archive_dir'] == 'none') { // don't archive anything
+            return true;
+        }
         $status = GetTestStatus($id, false);
         $completed = true;
         if ($status['statusCode'] >= 100 && $status['statusCode'] < 200)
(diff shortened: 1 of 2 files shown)


www/archive.inc
```diff
@@ -17,6 +17,9 @@ function ArchiveTest($id, $store_result = true, $delete = false) {
     if ( strpos($id, '.') === false &&                           // make sure it isn't a relay test
         (array_key_exists('archive_dir', $settings) ||           // local mount
         array_key_exists('archive_s3_server', $settings)) ) {    // S3 Interface (Internet Archive or standard)
+        if ($settings['archive_dir'] == 'none') { // don't archive anything
```
pmeenan: Can you make this a case-insensitive comparison just to be safe (and potentially also allow for /dev/null)?

Something like:
```
if (strcasecmp($settings['archive_dir'], 'none') === false || strcasecmp($settings['archive_dir'], '/dev/null') === false) {
```

pmeenan: Err, make that:

```
if (!strcasecmp($settings['archive_dir'], 'none') || !strcasecmp($settings['archive_dir'], '/dev/null')) {
```

www/cli/archive.php
```diff
@@ -225,7 +225,10 @@ function CheckTest($testPath, $id, $elapsedDays) {
     if (isset($elapsed)) {
       if( $elapsed >= $MIN_DAYS ) {
           $delete = true;
-          if (ArchiveTest($id) ) {
+          if ($archive_dir == 'none') { // archive_dir == none indicates don't archive anything
```
pmeenan: This shouldn't be needed since ArchiveTest() already does the check

pmeenan:
Closing from lack of activity.  Please re-open if you take another run at it.

PR #414, opened 2015-03-12
Added ability to set timeout value on per-test basis

Kevin-Jonaitis:
This allows users to set a query parameter "timeout=%d" in runtest.php,
where %d is the amount of time to test should be allowed to run for
before web page test force quits the test.

The results of this test up to this point of timeout will be recorded.
If the test times out, the result code for the step that that timeout
occurred on will be 99998.

Manual Testing:
-Verified new code has no effect when EnableUserSetTimeout is not included
in wptdriver.ini file
-Verified test times out when timit limit is reached
-Verified test does not timeout when time limit is not reached
-Verified negative timeout values are discarded
-Verified that partial test results are returned on multi-step test run

### DIFF
--- a/agent/wptdriver/web_browser.cc
+++ b/agent/wptdriver/web_browser.cc
@@ -232,7 +232,7 @@ bool WebBrowser::RunAndWait() {
           ret = true;
           _status.Set(_T("Waiting up to %d seconds for the test to complete"), 
                       (_test._test_timeout / SECONDS_TO_MS) * 2);
-          DWORD wait_time = _test._test_timeout * 2;
+          DWORD wait_time = _test._test_timeout; //Stricter limit on this value 
           #ifdef DEBUG
           wait_time = INFINITE;
           #endif
--- a/agent/wptdriver/wpt_test.cc
+++ b/agent/wptdriver/wpt_test.cc
@@ -33,7 +33,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "../wpthook/shared_mem.h"
 #include "wpt_settings.h"
 
-static const DWORD SCRIPT_TIMEOUT_MULTIPLIER = 2;
 static const BYTE JPEG_DEFAULT_QUALITY = 30;
 static const DWORD MS_IN_SEC = 1000;
 static const DWORD BROWSER_WIDTH = 1024;
@@ -59,7 +58,9 @@ WptTest::WptTest(void):
   ,_activity_timeout(DEFAULT_ACTIVITY_TIMEOUT)
   ,_measurement_timeout(DEFAULT_TEST_TIMEOUT)
   ,has_gpu_(false)
-  ,lock_count_(0) {
+  ,lock_count_(0),
+  _script_timeout_multiplier(2),
+  _enableUserSetTimeout(false) {
   QueryPerformanceFrequency(&_perf_frequency);
 
   // figure out what our working diriectory is
@@ -157,6 +158,7 @@ void WptTest::Reset(void) {
   _run_error.Empty();
   _test_error.Empty();
   _custom_metrics.Empty();
+  _has_test_timed_out = false;
 }
 
 /*-----------------------------------------------------------------------------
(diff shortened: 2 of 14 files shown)


agent/wptdriver/web_browser.cc
```diff
@@ -232,7 +232,12 @@ bool WebBrowser::RunAndWait() {
           ret = true;
           _status.Set(_T("Waiting up to %d seconds for the test to complete"), 
                       (_test._test_timeout / SECONDS_TO_MS) * 2);
-          DWORD wait_time = _test._test_timeout * 2;
+          DWORD wait_time;
+          if (_test._overall_timeout_in_ms != 0 &&
+            _test._overall_timeout_in_ms < _test._test_timeout * 2)
+            wait_time = _test._overall_timeout_in_ms; //Strict limit on this value
+          else
+            wait_time = _test._test_timeout * 2;
```
pmeenan: Why not keep using _test_timeout as the overall timeout?  *2 is probably a bit much but adding 30-60 seconds on whatever it is set to to allow for browser startup and data collection should be reasonable.

agent/wptdriver/wpt_test.cc
```diff
@@ -286,8 +288,13 @@ bool WptTest::Load(CString& test) {
         else if (!key.CompareNoCase(_T("addCmdLine")))
           _browser_additional_command_line = value;
         else if (!key.CompareNoCase(_T("continuousVideo")) &&
-                 _ttoi(value.Trim()))
+          _ttoi(value.Trim()))
           _continuous_video = true;
+        else if (!key.CompareNoCase(_T("timeout"))) {
+          _overall_timeout_in_ms = _ttoi(value.Trim()) * 1000;
```
pmeenan: Why a new property and not just use _test_timeout?  It is initialized to a default value but that doesn't mean it couldn't/shouldn't be overridden.

agent/wpthook/test_state.cc
```diff
@@ -319,7 +325,14 @@ bool TestState::IsDone() {
   if (_active) {
     bool is_page_done = false;
     CString done_reason;
-    if (test_ms >= _test._minimum_duration) {
+    DWORD elapsed_timeout_ms = ElapsedMsFromTimeoutStart(now);
+    if (_test._overall_timeout_in_ms > 0 &&
+      elapsed_timeout_ms > _test._overall_timeout_in_ms) {
```
pmeenan: This should be comparing the _timeout_start_time, not _step_start since it is checking for an overall timeout, right?

agent/wpthook/test_state.cc
```diff
@@ -319,7 +325,14 @@ bool TestState::IsDone() {
   if (_active) {
     bool is_page_done = false;
     CString done_reason;
-    if (test_ms >= _test._minimum_duration) {
+    DWORD elapsed_timeout_ms = ElapsedMsFromTimeoutStart(now);
```
pmeenan: Don't really need a whole new method for this, just call ElapsedMs directly with _timeout_start_time

pmeenan:
Just a couple of nits and one main concern - it looks like it allows arbitrary user timeouts which would be bad (say I set a timeout of 10 years).  Can you add logic to runtest.php to check for a max timeout setting (or use a default of something like 10 minutes if not specified)?

agent/wptdriver/web_browser.cc
Kevin-Jonaitis: The stricter this limit is, the better. We want WPT to run to as close to the timeout value  as possible. (obviously this isn't going to be perfect as there's a boot-up time for the browser and shutdown time). There's also a 10 second wait time after this statement for the browser process to finish, so I think that's more than adequate to let the browser finish saving/shutdown.

agent/wptdriver/wpt_test.cc
Kevin-Jonaitis: _test_timeout is actually used for a per-step basis, and is reset after every step. Since we want to keep track of the overall test time(including time spent between steps), I created a new field.

agent/wpthook/test_state.cc
Kevin-Jonaitis: Your right, removed

Kevin-Jonaitis:
You're right. I originally had a field in the .ini file called EnablePerTestTimeout. I will re-add it. It basically does the following:

Setting EnablePerTestTimeout=1 allows users to enable to set query parameter timeout=%d in their runtest.php file. To prevent people from setting an arbitrarily large timeout on public WPT instances, the instances can simply not include this line or set it to EnablePerTestTimeout=0. Instances that DO enable this feature should be weary of users setting inappropriately large timeout values, as there will be no upper limit.

pmeenan:
Sorry about the delay - LGTM.

PR #422, opened 2015-04-03
Capture Chrome console messages and send to wptdriver

andydavies:
Tested against
https://dl.dropboxusercontent.com/u/91644/tests/console/index.html

Supports .assert, .debug .log .info, .warn, .error

Also passes .data for  .table but WPT UI doesn't display it currently

Although the extension grabs the data for .dig by the time it's stored
in 1_console_log.json.gz some of the object detail is lost so it's not
displayed correctly either


andydavies:
I woke up this morning an realised I should probably create a ```wpt.chromeDebugger.SendConsoleData``` method so I'll redo this PR a bit later

andydavies:
Actually looking at the rest of the method there's code like ```wpt.chromeDebugger.sendEvent('trace', JSON.stringify(params['value']));``` so perhaps I don't need to encapsulate in a new method.

LMK what convention you'd prefer and I can follow that!

PR #427, opened 2015-04-16
Add support for full size video capture

ashvince:
This adds the ability to capture all progress images that make up the video at full resolution even if this results in the images that are created being greater than 600 pixels wide.

Option can be disabled or defaulted on or off via settings file.


pmeenan:
Very clean PR - appreciate it.

PR #432, opened 2015-04-26
Resolve IPs and test to make sure they are not in a reserved range

kjoconnor:
I haven't written PHP in a very long time so let me know if this doesn't match any style guidelines. I just tried to match the functions around it.

I had emailed you about this privately and you ended up adding the link local checks about a month ago (thanks!), I think this just makes doubly sure that we wouldn't be requesting strange pages.

### DIFF
--- a/www/runtest.php
+++ b/www/runtest.php
@@ -1362,10 +1362,65 @@ function ValidateURL(&$url, &$error, &$settings)
         $error = "Sorry, $host is blocked from testing";
     elseif( !strcasecmp(substr($url, -4), '.pdf') )
         $error = "You can not test PDF files with WebPagetest";
-    else
+
+    $ip_address = gethostbyname($host);
+
+    if (ValidateIP($ip_address, $settings) !== FALSE) {
         $ret = true;
+    } else {
+        $error = "You can not test <b>$ip_address</b> from the public Internet.  Your web site needs to be hosted on the public Internet for testing";
+    }
 
     return $ret;
+}
+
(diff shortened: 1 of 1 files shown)


pmeenan:
Only change I'd suggest is to only go the gethostbyname if private IP testing is disabled.

That said, it would probably be better to pass a flag down to the test agents that indicates private testing is not allowed and have them fail the connections like we do for the link-local addresses.  Otherwise there are WAY too many ways around it (test a page that redirects to a private IP, test a page with an iFrame to the private IP, etc).

kjoconnor:
Do you know whereabouts in the code that flag could be passed? Or would that require a major rearchitecting?

pmeenan:
The flag itself would be passed in runtest.php when the task is built -
https://github.com/WPO-Foundation/webpagetest/blob/master/www/runtest.php#L1806
(probably something like "blockprivate=1").

Then in wptdriver where the connections are hooked we'd check the flag and
if the IP was a private IP like we check link local:
https://github.com/WPO-Foundation/webpagetest/blob/master/agent/wpthook/track_sockets.cc#L117

Same in urlblast -
https://github.com/WPO-Foundation/webpagetest/blob/master/agent/browser/ie/pagetest/WsHook.cpp#L264

Most of the work would probably be in plumbing the setting through from the
task into the parts of the code that do the blocking.

pmeenan:
Closing from lack of activity.  Please re-open if you take another run at it.

PR #444, opened 2015-05-14
Add Surge as a known CDN

JakeChampion:
[Surge](https://surge.sh/) is a static hosting platform which also has its own CDN platform.


pmeenan:
Will go live on the public instance when the agent is next updated (usually a week or two).

PR #451, opened 2015-06-16
Support basic authentication and HTTPS when communicating with WP

arolleston:
Support basic authentication and HTTPS when communicating with the WPT server.  Can also select whether a valid certificate is expected.  All parameters are configurable through the settings.ini or EC2 user-data.

Note that the credentials are sent with the initial request to the WPT server, instead of sending a request without credentials, detecting the 401 response, then re-requesting with the credentials.

### DIFF
--- a/agent/wptdriver/webpagetest.cc
+++ b/agent/wptdriver/webpagetest.cc
@@ -537,11 +537,11 @@ bool WebPagetest::CrackUrl(CString url, CString &host, unsigned short &port,
       object = path;
       object += extra;
       if (!lstrcmpi(scheme, _T("https"))) {
-		    secure_flag = INTERNET_FLAG_SECURE;
-		    if (!_settings._requireValidCertificate) {
-			    secure_flag |= INTERNET_FLAG_IGNORE_CERT_CN_INVALID |
-				    INTERNET_FLAG_IGNORE_CERT_DATE_INVALID;
-		    }
+        secure_flag = INTERNET_FLAG_SECURE;
+        if (!_settings._requireValidCertificate) {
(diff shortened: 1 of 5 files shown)


agent/wptdriver/webpagetest.cc
```diff
@@ -283,6 +282,18 @@ bool WebPagetest::UploadData(WptTestDriver& test, bool done) {
 }
 
 /*-----------------------------------------------------------------------------
+  Set the credentials required to access the server, if configured
+-----------------------------------------------------------------------------*/
+void WebPagetest::SetLoginCredentials(HINTERNET request) {
+  if (!_settings._username.IsEmpty() & !_settings._password.IsEmpty()) {
```
pmeenan: This should be &&

pmeenan:
Just the one minor change and it looks ready to go - thanks for doing this.

arolleston:
Fixed the ampersand.  Add new settings and documented them in wptdriver.ini.sample.

Documentation of the new EC2 user data parameters for the [Private Instances](https://sites.google.com/a/webpagetest.org/docs/private-instances#TOC-Parameters) page
> wpt_username - username for basic authentication with WPT server. Ignored if wpt_password is not specified. (optional)
> wpt_password - password for basic authentication with WPT server. Ignored if wpt_username is not specified. (optional)
> wpt_validcertificate - Ignored unless the scheme of the wpt_server is 'https'  If the value is '0', the hostname and expiry of the certificate are not checked.  If the value is '1', the CN of the WPT server SSL certificate must match the hostname specified in the wpt_server parameter, and the certificate must be valid.  (optional, defaults to '0')

PR #461, opened 2015-07-16
added the ability to parse scripts into steps by navigate command

past5:
- mobile agent now parses navigate commands in user scripts into steps to be processed by wd_server
- wd_server is not yet changed and currently only processes the last navigate command


pmeenan:
Closing as the JS agent is deprecated and no longer supported (and the new agent supports multi-step scripts).

PR #478, opened 2015-09-02
Add Incapsula header detection

ZigZagIL:


pmeenan:
Can you also update agent/browser/ie/pagetest/cdn.h which has the definitions for the (deprecated) urlblast agent (note the strings are slightly different).

PR #497, opened 2015-10-03
Add GoCache as a known CDN

gocachebr:
GoCache is a next-gen CDN platform to protect and speed up any website in a smart and simple way.


pmeenan:
Do you have an example site running on GoCache or a network map of some kind showing POPs?  I just need to check to make sure it isn't just offloading static files to separate hosting and actually distributes content to various locations and serves from the location closest to the user (the primary reason for WPT's CDN detection).

PR #499, opened 2015-10-06
Add Ensighten tag manager as CDN

GlynnPhillips:
When testing http://www.nature.com/nplants/ recently I noticed Ensighten
tag manager was not within the list of CDN's.

As far as I am aware Ensighten are distributing content to different
locations and serving from the relevant one.

Slightly out of date but here is a map of Ensightens global network
https://www.ensighten.com/sites/default/files/Ensighten-Tag-Delivery-Network.pdf

### DIFF
--- a/agent/browser/ie/pagetest/cdn.h
+++ b/agent/browser/ie/pagetest/cdn.h
@@ -103,6 +103,7 @@ CDN_PROVIDER cdnList[] = {
 	{".rlcdn.com", _T("Reapleaf")},
 	{".wp.com", _T("WordPress Jetpack")},
 	{".incapdns.net", _T("Incapsula"}),
+	{"nexus.ensighten.com", _T("Ensighten"}),
 	{NULL, NULL}
 };
 
--- a/agent/wpthook/cdn.h
+++ b/agent/wpthook/cdn.h
@@ -142,6 +142,7 @@ CDN_PROVIDER cdnList[] = {
   {".cdninstagram.com", "Facebook"},
   {".rlcdn.com", "Reapleaf"},
   {".wp.com", "WordPress Jetpack"},
+  {"nexus.ensighten.com", _T("Ensighten"}),
   {"END_MARKER", "END_MARKER"}
 };
 


andydavies:
I've been looking at Ensighten for a few clients lately and using dig in
the UK the ensighten domains appear to resolve to an ELB on US-West-1
rather than be fronted by a CDN

There's a separate question as to whether a tag manager should be marked as
a CDN my view is they shouldn't but if they use a CDN e.g. Cloudfront,
Akamai etc. that should be apparent in the test

GlynnPhillips:
Thanks @andydavies. Could you show me how you have been using the dig command to look this up, so I could run similar test? I have tried various commands and tools but I am certainly no expert in that area.

I was also questioning whether tag managers should be listed as a CDN as initially I felt that the same, that they shouldn't. However then I started to think that this comes down to how you define a CDN and the more I thought about this I thought some tag managers might be CDN's as well. This would come down to the tag manager serving the content from a relevant location mind.

GlynnPhillips:
Actually I guess you were just running `dig nexus.ensighten.com` in which case I am also getting a US location.

I guess this would highlight they are not like a CDN so this could be closed.

pmeenan:
+1  I did a few traceroutes from various locations and I'm seeing 70-120ms RTT's and also going to a S3 ELB.  They may be using DNS for emergency failover or the logic may be configured for their best costs rather than performance but it's not doing geo-location serving for performance (which is a real shame since they are usually a blocking script at the top of a page).

GlynnPhillips:
Yeah we ran a few more tests and looks like we sometimes get EU locations and then sometimes get US locations, we are London based. So it looks like they are too regularly falling back to US locations.

We are going to speak with our contact for our own piece of mind to understand what process they are using.

andydavies:
So funnily enough after tweeting about their performance a couple weeks
ago,  I'm going in for a chat with them this afternoon!

Glyn: Are you going to the LDN perf meet up tonight?

GlynnPhillips:
Afraid not, I forgot about it whilst tickets were available.

PR #505, opened 2015-10-14
Enable export.php to produce multistep HARs

kart:
export.php now has enough information provided by the agent that it can generate multistep HARs. The rest of the UI has not changed, and is completely backward compatible with the older agents. Also, older servers can continue to serve newer agents. There are some features that are not adapted to be multistep -- like custom metrics, user timings, CSI metrics, etc. These will come gradually, but the most important results: basic page, and request details along with visual metrics for each step is now provided in the HAR. This will help tools that can manage their own UI to consume the new HARs and render multipage performance view.

More details can be found in the commit message.

Testing notes:
-- Several tests done with different variations to make sure multistep=0 and multistep=1 works. 
-- Verified that visualComplete, speed-index metrics are correctly computed.
-- Tested backward compatibility with older agent and newer server and vice-versa.

### DIFF
--- a/agent/wptdriver/util.cc
+++ b/agent/wptdriver/util.cc
@@ -44,6 +44,12 @@ static const TCHAR * DOCUMENT_WINDOW_CLASSES[] = {
   _T("WebKit2WebViewWindowClass")
 };
 
+// this should be maintained in sync with the capability bitmask constants in common_lib.inc
+// on the server side.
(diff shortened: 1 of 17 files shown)


kart:
@pmeenan for reference, you can take a look at a sample multistep HAR [here](https://github.com/Appdynamics/webpagetest/files/13544/multistep.151013_GX_4.txt)

PR #510, opened 2015-10-20
Expose 'mobile' status in XML results

josephscott:
When looking at XML test results there is nothing to indicate if the test was run as mobile or not. This patch adds 'mobile' to the XML results ( `<mobile>1</mobile>` ) when the test info indicates it was a mobile test.

### DIFF
--- a/www/xmlResult.php
+++ b/www/xmlResult.php
@@ -90,6 +90,8 @@ else
             }
             if( @strlen($test['testinfo']['mobile']) )
                 echo "<mobile>" . xml_entities($test['testinfo']['mobile']) .   "</mobile>\n";
+            if( isset($test['testinfo']['mobile']) )
+                echo "<mobile>" . xml_entities($test['testinfo']['mobile']) .   "</mobile>\n";
             if( @strlen($test['testinfo']['label']) )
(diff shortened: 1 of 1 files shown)


www/xmlResult.php
```diff
@@ -88,6 +88,8 @@
                 echo "<latency>{$test['testinfo']['latency']}</latency>\n";
                 echo "<plr>{$test['testinfo']['plr']}</plr>\n";
             }
+            if( @strlen($test['testinfo']['mobile']) )
```
pmeenan: The mobile flag is currently an integer, not a string: https://github.com/WPO-Foundation/webpagetest/blob/master/www/runtest.php#L170

Maybe just check isset and pass the value through?  That way it will cover it if I extend it to allow for multiple mobile profiles.

josephscott: I'm fine with that.  So make it:

```
            if( isset($test['testinfo']['mobile']) )
                echo "<mobile>" . xml_entities($test['testinfo']['mobile']) .   "</mobile>\n";
```

Works fine in my tests.

I think I may have messed up this pull request though.

pmeenan:
I'll clean it up after merging.

PR #556, opened 2016-01-25
Add/change CDNs

emilstahl:
wp.com is used for much for than Jetpack.


agent/wpthook/cdn.h
```diff
@@ -164,6 +164,11 @@ CDN_PROVIDER cdnList[] = {
   {".bisongrid.net", "Bison Grid"},
   {".cdn.gocache.net", "GoCache"},
   {".hiberniacdn.com", "HiberniaCDN"},
+  {".cdntel.net", "Telenor"},
+  {".raxcdn.com", "Rackspace"},
+  {".keycdn.com", "KeyCDN"},
```
lucasRolff: They only announce zones on `kxcdn.com`

PR #592, opened 2016-04-01
MozJPEG tip

pornel:
I've noticed previous tips were missing target=_blank, so I've fixed that too.


pmeenan:
Awesome, thanks.  I've been debating moving away from the target=_blank model but thanks for fixing it.

Issue #625, opened 2016-05-24
Windows Agent: combineSteps is broken


sburnicki:
[combineSteps](https://sites.google.com/a/webpagetest.org/docs/using-webpagetest/scripting#TOC-combineSteps) in scripts does not work as documented.

Some scripts work as expected:

```
combinesteps
navigate    http://www.google.com
navigate    http://www.aol.com
```

Whereas it doesn't seem to work with a specified number of steps to combine (only aol.com is being measured/appears in the waterfall):

```
combinesteps    2
navigate    http://www.google.com
navigate    http://www.aol.com
```

With a logData block in before, no data is saved and sent by the agent at all:

```
logData 0
navigate    http://www.yahoo.com
logData 1
combinesteps
navigate    http://www.google.com
navigate    http://www.aol.com
```

It seems to work with only a single step after `logData 1`, though, so it's definitely related to the appearing of combinesteps.

Issue #618, opened 2016-05-13
Multi-step measurement implementation


sburnicki:
With a multi-step implementation, we want to get multiple separated measurements by executing a single script. Currently, you can only get one result per executed script.

There is an existing fork with a multi-step implementation (https://github.com/iteratec-hh/webpagetest/tree/multistep), however, the real multi-step implementation is somehow lost in different commits and the fork is not compatible with upstream.

For a new, clean implementation of multi-step I'd like to discuss the implementation before actually starting with it.

**How is multi-step implemted in the fork?**
The desired behavior is to be downwards-compatible to the current single-step version.
Each new measurement starts therefore with a `setEventName` command.
Commands executed between two `setEventName` are regarded as a single measurement, which work equally to the current implementation.

A script with three measurements would therefore look like this:

    setEventName google
    navigate http://www.google.com
    setEventName aol
    navigate http://www.aol.com
    setEventName yahoo
    navigate http://www.yahoo.com

If we'd leave out the line `setEventName yahoo`, the second result of the measurement would be just as executing both navigates to aol and yahoo with the upstream WebPagetest version.

Another upside of this behavior is that there are very little changes to the current implementation of the script processing. The main differences to be made are in saving, sending them to the server, displaying the results, and make them accessible via REST API.

pmeenan:
I don't know that the script modifications are strictly necessary.  Even without them the agent code generates synthetic names and knows the step number it is currently working on.  It is certainly helpful when reporting or presenting to the users.

For me, the biggest requirement is for backward compatibility with both the agents and consumers of the API.  The structure of the JSON/XML/HAR can't change for the case of a single run and it would be best if the multi-step case was still backward compatible (possibly returning the combined data of all steps as a sequence where a single run would normally report).  There is a lot of tooling and automation that uses the existing API and it needs to continue to work.

Something like this might work:

- The "firstView" and "repeatView" entries under runs, average, standardDeviation and median would be the sum of the metrics from all of the steps.
- The median run would be selected based on the aggregate timings for a given run
- In multi-step cases only:
    - The requests would be missing and would need to be pulled from the individual steps
    - The video frames would be missing and would need to be pulled from the individual steps
    - There would be a "steps" array that contained the full "firstView"/"repeatView" entries for each step
    - Each step would include the label as well as the start time relative to the first step
    - The video and request timings in a given step would be relative to the start of that one step

[...]

PR #626, opened 2016-05-24
wptdriver: Adjustments for multistep support

sburnicki:
First changes to tackle multistep support as discussed in #618.
Does not fix chombineSteps as reported in #625.

The wptdriver basically supports measuring multiple steps, but the old implementation used the same file names for all steps to save.

Now, the first measured step uses the old file names to save the results, whereas all following steps get another prefix to identify the step. Therefore new prefixes starts from `_2` for step 2, etc.

This is backwards-compatible since only more files are being sent to the server, which simply ignores the new files. Executing a multi-step script with this version would result in only seeing the results for the first step.

### DIFF
--- a/agent/wpthook/results.cc
+++ b/agent/wpthook/results.cc
@@ -129,11 +129,7 @@ void Results::Save(void) {
   if (!_saved) {
     ProcessRequests();
     if (_test._log_data) {
-      reported_step_++;
-      if (_test._current_event_name.IsEmpty())
-        current_step_name_.Format("Step %d", reported_step_);
-      else
-        current_step_name_ = _test._current_event_name;
+      IncrementStep();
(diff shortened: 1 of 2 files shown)


pmeenan:
Looks good.  One tweak, can you have the bracket indents match the rest of the code (opening brace on the same line as the if)?  i.e.

```
if (blah) {
}
```

instead of 
```
if (blah)
{
}
```

The code itself looks good (though clearly parsing out the step numbers on the server is going to be a pain - ugh, backwards compatibility).

sburnicki:
Oops sorry, who knows why I chose the other coding style.
I amended the old commit to preserve a nice history.

And yes, parsing the steps will be ugly - but I guess sending separate step information and *validating them* on the server would result in even more effort.

PR #627, opened 2016-05-24
Add a script to purge old archived test results

anupdhml:
New cli script purgearchive.php that removes archived tests older than a month.

This is useful for saving on disk space. Noticed that there's a [pull request](https://github.com/WPO-Foundation/webpagetest/pull/392) open with similar goals, but that one bypasses archiving altogether. We have found it useful to archive tests that are few days old, and then from that archived set, this script deletes anything older than a month.

I did not add a configuration setting for the days limit here so as not to complicate `settings.ini`. 30 days seems reasonable and the other script `prunearchive.php` also uses the same value hardcoded.

### DIFF
--- a/www/cli/purgearchive.php
+++ b/www/cli/purgearchive.php
@@ -1,0 +1,84 @@
+<?php
+/* Based on prunearchive.php (which relies on archive2.php being run)
+ * This takes a simpler approach: just delete the day folders in archive_dir,
+ * if they are older than a month.
+ */
+
+chdir('..');
+
+include 'common.inc';
+ignore_user_abort(true);
+set_time_limit(86400);   // only allow it to run for 1 day
+
+// bail if we are already running
+$lock = Lock("Remove Archive");
+if (!isset($lock)) {
+  echo "removearchive process is already running\n";
+  exit(0);
+}
+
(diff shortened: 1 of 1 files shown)


www/cli/purgearchive.php
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+chdir('..');
+
+include 'common.inc';
+ignore_user_abort(true);
+set_time_limit(86400);   // only allow it to run for 1 day
+
+// bail if we are already running
+$lock = Lock("Remove Archive");
```
pmeenan: nit: can you make the lock name "purgearchive" so it is clear that it is tied to this script?

www/cli/purgearchive.php
```diff
@@ -0,0 +1,84 @@
...
+if (!isset($lock)) {
+  echo "removearchive process is already running\n";
```
pmeenan: And the name here (presumably it was renamed)

pmeenan:
Mostly good.  If you could rename the name of the lock I'll go ahead and merge it.

anupdhml:
@pmeenan yep forgot to update the names. Just updated it.

Issue #630, opened 2016-05-26
Agent 278 setViewportSize not working correctly in Chrome


zeman:
Since agent 278 we're seeing issues with the setViewportSize script command. In this test the viewport was set to 1280x800 but you can see that the webpage tested reports that the size of the browser viewport was 1306x683.
http://www.webpagetest.org/result/160526_C6_19PY/1/screen_shot/

This test page was used to report the viewport size of the test agent.
http://viewportsizes.com/mine/

pmeenan:
Is it possible that it was a change in Chrome (51 just came out) or can you reliably get 277 to work and 278 to fail?  Looking at the commit history there wasn't anything anywhere near code that would touch that between 277 and 278.  The only agent-impacting changes were around the UA string.

zeman:
Actually we skipped a version so we went from agent 276 to 278 if that makes a difference. Here's an example test set to 1280x800, both using Chrome 50.

Right viewport size using agent 276
https://wpt.speedcurve.com/screen_shot.php?test=160520_GD_753c724b1296dc22db6081ca410a8c77&run=1

Wrong viewport size using agent 278
https://wpt.speedcurve.com/screen_shot.php?test=160525_HN_e521afd33fd28735222232ada9d1c722&run=3&cached=0

If you open up the pngs you'll see they are very different sizes even though the script was identical.

arolleston:
I have also encountered this issue.  We have been running v271 since April 25 on a private instance.  Looking at our test history, we started seeing the problem on May 3, which is about the time that Chrome 50 was widely deployed.

Here are some other examples of the problem.  All tests had 9 runs
http://www.webpagetest.org/result/160531_QX_a5aecfced18c181c67405f7cfccfb8b4/
- Screenshot size for 8 runs was 361x364, run 4 was 381x480
 
http://www.webpagetest.org/result/160531_AA_77386fb791b26e8e76708f81883b67b8/
- Screenshot size for 8 runs was 361x364, run 8 was 381x815.

pmeenan:
Looks better: http://www.webpagetest.org/result/160604_ZB_FEX/

But from the looks of the screen shots here, there is still a bit of variability (in the width): http://www.webpagetest.org/result/160604_EF_d72091a086de0865c26e50aed0f0d96b/

zeman:
Certainly an improvement but seeing variations in the height rather than width. These should all be 1024 wide and 768 high.
[...]

PR #631, opened 2016-06-03
Fix Chrome viewport sizing by loading an empty document before attemp…

david-poirier-csn:
…ting resize

Fix Firefox viewport sizing by forcing viewport re-measurement inside UpdateBrowserWindow method
Fix off-by-one errors inside viewport measurement

#630

### DIFF
--- a/agent/wpthook/test_server.cc
+++ b/agent/wpthook/test_server.cc
@@ -299,9 +299,11 @@ void TestServer::MongooseCallback(enum mg_event event,
       SendResponse(conn, request_info, RESPONSE_OK, RESPONSE_OK_STR, "");
     } else if (strcmp(request_info->uri, "/event/received_data") == 0) {
       test_state_.received_data_ = true;
-    } else if (strncmp(request_info->uri, "/blank", 6) == 0) {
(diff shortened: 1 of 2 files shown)


zeman:
Does this fix the viewport sizing issues we're seeing here https://github.com/WPO-Foundation/webpagetest/issues/630

david-poirier-csn:
I reckon it should. The issues are probably caused by newer versions of Chrome creating child processes in a different way, but that's just a guess. The fix makes sure the window is open before attempting a resize, which "works on my machine" :)

Testing would be appreciated. You'd need to replace your "wptdriver.exe" and "wpthook.dll" files with ones built from my fork. If you need binaries let me know and I'll put them in dropbox for you.

pmeenan:
No need for a custom build.  Looks simple enough and other that curiosity why it is being done after the second blank page instead of the first, should be low risk.  I just pushed wptdriver 280 to prod with the fix.

@david-poirier-csn - huge thanks for looking into it and the fix.

david-poirier-csn:
No worries! 
I changed it to resize on the second load because when using Chrome it was trying to resize before the window was created, and the resize code will not try to operate on the same window twice. I could have changed that, but it seems simpler to load a blank page and do a no-op.

PR #632, opened 2016-06-06
Disable Windows 10 upgrade via Windows Update or nag dialog

david-poirier-csn:
These registry changes require a reboot to take effect


agent/wptdriver/wpt_driver_core.cc
```diff
@@ -819,6 +838,10 @@ void WptDriverCore::PreTest() {
       CloseHandle(process);
   }
 
+  // Minimise all windows on desktop
+  HWND hwnd = FindWindow(_T("Shell_TrayWnd"), NULL);
+  LRESULT res = SendMessage(hwnd, WM_COMMAND, (WPARAM)419, 0);
```
pmeenan: This is kind of hacky and not supported (though seems to be used frequently on stack overflow).  I'd feel better if this used GetDesktopWindow(), got the first child, enumerated all of the next windows and minimized each one like the dialog closing code: https://github.com/WPO-Foundation/webpagetest/blob/master/agent/wptdriver/wpt_driver_core.cc#L698

pmeenan:
Thanks.  Only tweak I'd request is to the window minimizing code to do it in a more-supported fashion.

david-poirier-csn:
Good pickup, that snuck in with a separate commit. I didn't realise Github would roll it into this pull request. Fixed.

PR #638, opened 2016-06-10
Server: Refactoring of XML result generation

sburnicki:
This is a prerequisite for upcoming changes related to #618.

I took the code to generate run-specific XML and encapsuled it as `XmlResultGenerator::printRun`.

Doing so, I used (and extended) the `TestPaths` class. and added a `UrlGenerator` which works for both friendly URLs (mod_rewrite) and standard URLs.
This made the function more dynamic, so it can be used for generating the XML of both first and repeat view.

Related functions were moved to the same file, as it makes sense to include them in the class in a future commit.

These changes make upcoming multi-step support much easier to implement, since most of the XML will be needed to be generated for each step, rather than each run.

The generated XML is **very** similar to the old XML.
There are some minor changes for the cached view of each run:

  - new-line between `<rawData>` and `<headers>` tag
  - missing link for `optimization_report` as it seems to be deprecated
    and points to a file which is not even included in the git project
  - added `/cached/` suffix for `breakdown` and `domains`. These URLs
    aren't necessary, but supported by current rewrite rules and the
    target files.

I locally downloaded a xmlResult from my wpt test server running a master branch and again with my new changes. The diff can be seen here: http://pastebin.com/1JR1A9Cn


pmeenan:
Looks awesome, thanks.

PR #642, opened 2016-06-16
xmlResult: Further refactoring and tests

sburnicki:
I continued refactoring the xmlResult generation.
New classes:
* `FileHandler` to abstract from file system
* `TestInfo` to represent test information
* `TestRunResult` to represent the result of one run (cached or uncached)

The advantage of using these classes is that they can be mocked in tests and can be easily modified without changing their interface for XML generation.

`xml*` functions were moved into the `XmlResultGenerator` class, so they can take advantage of the new classes.

After all, XML output for test runs and median run doesn't depend on the same data.
Both XML for median and runs are tested via tests, so we can make sure it stays compatible in future changes (as #618).

The output of `xmlResult.php` is *exactly* the same as before.

Hopefully the different commits help you to understand the changes, as the combined diff looks quite huge.


bryongloden:
🎉

PR #647, opened 2016-06-23
Added akamaized.net CDN domain for Akamai.

juanluisbaptiste:
Hi,

There's a missing Akamai hostname, akamaized.net, I work for an Akamai reseller and we have some customers using that domain.


juanluisbaptiste:
Thanks.

Issue #653, opened 2016-07-06
Design of multistep result pages


sburnicki:
I want to start to implement multistep support (#618) in the different result pages.
Before getting started, I'd like to discuss the way the results hould be presented.

With multistep, the results per page increase. The way we chose in our fork before was to to just display all the results of all steps below each other (per run/view).

![image](https://cloud.githubusercontent.com/assets/5701519/16618431/98e82bbc-4389-11e6-93ef-c63af96582cf.png)

For some pages, we provided quicklinks on the top for better navigation on the page.

![image](https://cloud.githubusercontent.com/assets/5701519/16618462/ce07ebf2-4389-11e6-8818-87b3ad2ff753.png)

On the `details` page we also show all step results per run, but grouped (waterfalls, connection views, request details). This way it's for example possible to quickly compare waterfalls of different steps.
We introduced an "accordeon" to load the different results asynchronously and avoid long loading times.

![image](https://cloud.githubusercontent.com/assets/5701519/16618525/274d9432-438a-11e6-8bb8-c331f73bbdfe.png)

I'm also open for new ideas to make the results overseeable. If nobody has better ideas, I'd implement the pages similar to our fork.

pmeenan:
Sounds good to me.  Only piece I'm not sure about are the grades.  I wouldn't want multiple rows of grades all in the header because it would push all of the other content out of view.  Maybe average all of the grades for the one at the top and then include the per-step gradings with the individual results (or only on the details page to avoid over-cluttering the main results page).

pmeenan:
Not sure but it might be nice to auto-expand the first waterfall accordion.

Also, the quicklinks for the "customize waterfall" and "view all images" should probably just go to the actual destination pages since they are just links.

sburnicki:
Average grades in the header make sense. The first screenshot is actually from the performance review page, where grades for all steps are presented.

Regarding the waterfalls: I also think it makes sense to auto-expand the first waterfall. With step-specific anchor links (which are also used in quicklinks), the target accordeon also gets automatically expanded.

And I agree that quicklinks for "customize waterfall" and "view all images" should be the "real" links.

PR #659, opened 2016-07-13
Multistep changes for Screen Shot page and header

sburnicki:
First multistep changes for the UI as discussed in #653.
Screenshots/console log/status log of all pages are shown, grouped by step.
A quciklink table and navigation with `j` and `k` keys make it possible to navigate through the steps.

Also the header (used in all pages) was updated. Adult site detection, grades and the shown URL are now based on all steps, not only on the first. The grades are averages of all steps.

(Part of multistep implementation as discussion in #618.)

### DIFF
--- a/www/include/TestInfo.php
+++ b/www/include/TestInfo.php
@@ -135,4 +135,21 @@ class TestInfo {
       $tester = $this->rawData['testinfo']['test_runs'][$run]['tester'];
     return $tester;
   }
+
+  /**
+   * @param string[] $keywords Keywords to check the info for
+   * @return True if the TestInfo matches the keywords, false otherwise
+   */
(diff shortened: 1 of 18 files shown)


www/include/TestInfo.php
```diff
@@ -135,4 +135,21 @@ public function getTester($run) {
       $tester = $this->rawData['testinfo']['test_runs'][$run]['tester'];
     return $tester;
   }
+
+  /**
+   * @param string[] $keywords Keywords to check the info for
+   * @return True if the TestInfo matches the keywords, false otherwise
+   */
+  public function isAdultSite($keywords) {
+    $url = $this->getUrl();
```
pmeenan: The adult site check also checks the title (in addition to the URL)

PR #674, opened 2016-07-29
Add control and experiment urls to ab_analysis.py

csharrison:
Seems like we lost chromeUserTiming, so I'm removing ParseTime. I also cleaned up the CSV generation to add a header.


utils/analysis/ab_analysis.py
```diff
@@ -30,11 +30,10 @@
 
 # Set of metrics we are interested in, along with a function to extract that
 # metric from a WPT result row.
+# TODO(csharrison): Add ParseTime when we have valid start/end parse times.
 _METRICS = {
     'SpeedIndex': lambda x: sanitizeInt(x['SpeedIndex']),
     'TTFP': lambda x: sanitizeInt(x['render']),
-    'ParseTime': lambda x: sanitizeIntDiff(
-        x['chromeUserTiming.domInteractive'], x['chromeUserTiming.domLoading']),
```
pmeenan: User timing is still available but you need to explicitly enable tracing and blink.user_timing as a trace event (Chrome would crash if pages had invalid user timing marks which made them untestable with the trace event on by default).

What will probably work better though is if I add domLoading to the reported metrics directly from navigation timing since I'm already pulling domInteractive.

pmeenan: I just pushed the domLoading stand-alone metric so "domInteractive" and "domLoading" should both be available from the metrics directly without the need for user timing.

csharrison: Thanks, removed "chromeUserTiming" here.

PR #683, opened 2016-08-13
Add allowNonFQDN setting, useful for corporate intranets

NatGuyton:
When setting up WPT Private Instance on our company intranet, I had to make this modification to allow testing sites that only used DNS short names and were not set up for fully qualified names. 

This off-by-default setting allows a user to use a non-fully-qualified hostname in the URL.

### DIFF
--- a/www/runtest.php
+++ b/www/runtest.php
@@ -1448,7 +1448,7 @@ function ValidateURL(&$url, &$error, &$settings)
 
     if( strpos($url, ' ') !== FALSE || strpos($url, '>') !== FALSE || strpos($url, '<') !== FALSE )
         $error = "Please enter a Valid URL.  <b>" . htmlspecialchars($url) . "</b> is not a valid URL";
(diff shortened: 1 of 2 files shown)


www/runtest.php
```diff
@@ -1448,7 +1448,7 @@ function ValidateURL(&$url, &$error, &$settings)
 
     if( strpos($url, ' ') !== FALSE || strpos($url, '>') !== FALSE || strpos($url, '<') !== FALSE )
         $error = "Please enter a Valid URL.  <b>" . htmlspecialchars($url) . "</b> is not a valid URL";
-    elseif( strpos($host, '.') === FALSE )
+    elseif( strpos($host, '.') === FALSE && !$settings['allowNonFQDN'] )
```
pmeenan: Could you change this to GetSetting('allowNonFQDN') instead?

NatGuyton:
Updated, thanks!

PR #691, opened 2016-08-18
Added a hook to initiate custom result postprocessing

sburnicki:
In our private instance we have some very test specific result post processing. As it wouldn't make much sense to directly introduce these special cases in the main project, I'd like to introduce a way to include a custom hook.

So with these changes it's possible to add a `custom/hooks/precompletePostprocess.php` file gets included just before the test is marked as completed.

### DIFF
--- a/www/work/workdone.php
+++ b/www/work/workdone.php
@@ -325,6 +325,10 @@ if (ValidateTestId($id)) {
 
           // do pre-complete post-processing
           MoveVideoFiles($testPath);
+
+          if (file_exists('custom/hooks/precompletePostprocess.php')) {
+            include 'custom/hooks/precompletePostprocess.php';
+          }
           
           if (!isset($pageData))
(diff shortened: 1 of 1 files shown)


pmeenan:
If possible, could you try implementing it with the [plugins architecture](https://github.com/WPO-Foundation/webpagetest/blob/master/www/plugins.php.inc) that is sort of in place?  It's not documented at all but basically any .inc file in the $plugins folder will get loaded (I think it is pulled in by common.inc).  They just need to call WptAddHook() with the hook name and the function to call.

Then in the post processing you could call something like WptHookPostProcessTest() which would call all of the attached plugins.

sburnicki:
Oh yes of course, I wasn't aware of this possibility. This of course an even cleaner solution.

sburnicki:
Discarded in favor of #693 .

Issue #690, opened 2016-08-18
Problem with waterfall images & video for relayLocation


soulgalore:
Hey Pat,
I've been trying to setup webpagetest.org as an relayLocation. I can submit a job and get the result back, I can see that we have some data locally on our server and test start page works (or have some data) but images/video and some other pages doesn't work. This is an example: http://wpt.wmftest.org/result/160818_DT_5X/

Our server is on AWS and I added:
<pre>
[locations]
...
10=Dulles

[Dulles]
1=WPT-Dulles
label="Dulles"

....
[WPT-Dulles]
browser=Chrome,Firefox
label="WebPageTest.org Dulles"
relayServer="https://www.webpagetest.org/"
relayKey="OUR_SUPER_SECRET_KEY"
relayLocation="Dulles"
</pre>
in var/www/webpagetest/www/settings/ec2_locations.ini

I haven't found any errors in the logs but could be I'm looking in the wrong place, can you show me where to look?

Thanks Peter

sburnicki:
I guess this is an issue with the `UrlGenerator` I introduced to generate links.
The URL of the waterfall thumbnail is generated as `http://wpt.wmftest.org/result/ae622160747643b8bc288dbfc35bfa83.160818_DT_5X/1_waterfall_thumb.png`, but should be `http://wpt.wmftest.org/result/160818_DT_5X/1_waterfall_thumb.png`.

Can somebody tell me at which part the `ae622160747643b8bc288dbfc35bfa83` gets important? So I can keep it in mind when fixing the issue :)

soulgalore:
Sweet: waterfall and screenshots work now, thanks @sburnicki 

Filmstrip view and video do not work http://wpt.wmftest.org/result/160819_K8_3F/ (try clicking on the links).

sburnicki:
Similar issue, will fix later it later this day :)

pmeenan:
Looks like the test ID's were incorrectly getting changed on the relay server to include the relay key.  I just pushed the fix to the public server and did some quick testing and it appears to be working.  Can you let me know if it fixed the issues you were seeing.

Sorry if I'm a bit slow to respond, I'm out for the next week or so and working on a tiny laptop screen is....challenging ;-)

soulgalore:
Yes filmstrip and video works now, thanks :)

>Sorry if I'm a bit slow to respond

Np at all. You get better support from you than any paid product.

PR #692, opened 2016-08-18
Fix UrlGenerator for tests with relay IDs

sburnicki:
This is a fix for #690.

For friendly URLs, the relay key should be omitted in URLs.

However, I'm not sure if it's okay to simply omit the relay key. Manually using URLs without the relay key work on the test server in the issue, but I'm not sure if this is sufficient as `GetTestPath` seems to return a different path for IDs with relay key.

### DIFF
--- a/www/include/UrlGenerator.php
+++ b/www/include/UrlGenerator.php
@@ -10,6 +10,7 @@ abstract class UrlGenerator {
   protected $step;
   protected $baseUrl;
   protected $testId;
+  protected $testIdWithoutRelayKey;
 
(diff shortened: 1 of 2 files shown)


pmeenan:
The relay key generates a different path on the relay server but once the test is local it shouldn't be using the key.  This should work fine for now but there's probably a cleaner way to do it without so much duplication.

sburnicki:
So I guess we can completely remove the relay key from the ID in the UrlGenerator? It still causes some issue.

PR #701, opened 2016-08-26
Add label to bulk test if specified in UI

andydavies:
If a label is specified in UI, adds it to individual tests within bulk test so long as they don't have labels already specified.

Where individual tests already have labels specified these take precedence.

Code uses braces even though it's a single line if, aware it's different to rest of code base so  happy to remove them if you prefer.

### DIFF
--- a/www/runtest.php
+++ b/www/runtest.php
@@ -708,7 +708,7 @@
                         foreach( $bulk['urls'] as &$entry )
                         {
                             $testData = $test;
-                            if( $entry['l'] )
+                            if (isset($entry['l']) && strlen($entry['l']))
                             {
                                 $testData['label'] = $entry['l'];
                             }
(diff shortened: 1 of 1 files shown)


www/runtest.php
```diff
@@ -708,7 +708,10 @@ function DealWithMagicQuotes(&$arr) {
                         foreach( $bulk['urls'] as &$entry )
                         {
                             $testData = $test;
-                            $testData['label'] = $entry['l'];
+                            if( $entry['l'] )
```
pmeenan: Not sure if I know php well enough but I might feel better if this was:
```
if (isset($entry['l']) && strlen($entry['l']))
```

andydavies:
Your php knowledge is undoubtably better than mine so happy to change it (Originally I just followed the pattern in the rest of that section)

Issue #724, opened 2016-09-26
WPTLoader fails with an Access Violation exception with Firefox 48


Etenil:
When running a test on Firefox 48 with wptdriver 314 or the current master (8b9acd42b3b6b0db19597b37c213bce2f0499d44), wptload.dll fails at dllmain.cc:47 with an Access Violation exception; this is the place where the code attempts to read the user's name.

pmeenan:
I can't reproduce it here but either way it should be fixed with this: https://github.com/WPO-Foundation/webpagetest/commit/1f9c221d4623da8ffd4faf873b09c26fafd4866c

I'll roll out 315 in a little bit (have a few other things I want to get into it) and update when it is available.

pmeenan:
I just pushed out the 315 build of wptdriver which includes the fix (if you want pre-built binaries)

green-mdoerflinger:
@pmeenan Where can i get the latest release? Are you planning on releasing a private instance package in the near future?

pmeenan:
The agents can [update independently of the web UI](https://sites.google.com/a/webpagetest.org/docs/private-instances#TOC-Updating-Test-Agents) and are released regularly.

The latest release of the web UI is pretty ancient at this point and I'll be cutting a new one in a few weeks.  I usually recommend to just run off of the tip of master which is what the public instance does which also makes it easy to sync updates.

green-mdoerflinger:
Thanks!

PR #725, opened 2016-09-26
wptloader: Move username check to wpthook

Etenil:
This is an attempt to fix Issue #724 where the checking the user's name in wptloader is causing a memory access violation exception with Firefox 48 on Windows 7 32bits.

I haven't figured out why this exception is caused, and why it doesn't happen with other browsers. However I have moved the username check from `wptloader` to the very beginning of `wpthook` and this doesn't fail.

### DIFF
--- a/agent/wpthook/dllmain.cc
+++ b/agent/wpthook/dllmain.cc
@@ -29,6 +29,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // dllmain.cpp : Defines the entry point for the DLL application.
 #include "stdafx.h"
 #include "wpthook.h"
+#include <TCHAR.H>
+#include <Lmcons.h>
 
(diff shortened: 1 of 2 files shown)


pmeenan:
I think I'll have to find another way to early-bail.  There were issues with some system services not reacting well to loading wpthook.dll for some reason so I need an early, light-weight bail in the loader.  FWIW, it was only an issue in 64-bit so at a minimum, I can gate the check to only run in that case.

pmeenan:
I switched to use an explicit environment variable passed through wptdriver so the check will be a lot more resilient and only relies on kernel32 entry points: https://github.com/WPO-Foundation/webpagetest/commit/1f9c221d4623da8ffd4faf873b09c26fafd4866c

pmeenan:
Updated binaries should be available within a day or two if you need it and don't want to build from source

Etenil:
@pmeenan Great thanks a lot for this!

PR #726, opened 2016-09-27
Update cdn.h

namyongkim:
update KINX CDN

### DIFF
--- a/agent/wpthook/cdn.h
+++ b/agent/wpthook/cdn.h
@@ -171,6 +171,7 @@ CDN_PROVIDER cdnList[] = {
   {".raxcdn.com", "Rackspace"},
   {".unicorncdn.net", "UnicornCDN"},
   {".optimalcdn.com", "Optimal CDN"},
+  {".kinxcdn.", "KINX CDN"},
   {"END_MARKER", "END_MARKER"}
 };
 


pmeenan:
Are there a fixed set of tld's that people cname to and do they always cname?  If so I'd prefer to list all variants explicitly.

Also, any chance there is also a HTTP header (server, via, x-something)?  That would help with reliability in case the DNS lookup isn't captured.

namyongkim:
Yes, we service and which entrusted the CNAME CDN
ex) abcd.kinx.com, cdef.kinxcdn.net
Applicable?
--------------------------------------
https://www.webpagetest.org/result/160922_EH_7Q0/1/performance_optimization/#use_of_cdn
Glossary > Use A CDN 
Checked to see if it is hosted on a known CDN (CNAME mapped to a known CDN network). 80% of the static resources need to be served from a CDN for the overall page to be considered using a CDN. The current list of known CDN's is here

pmeenan:
Can you change the PR to have 2 separate entries, one for .com and one for .net (see the facebook example earlier in the list)?

pmeenan:
I took care of it here: https://github.com/WPO-Foundation/webpagetest/commit/cb26ce9df7f2414f833687edf00c9050f11c8e4e

It will go live on the public instance in the next 24 hours.

PR #731, opened 2016-10-06
Ensure the lower of two middle values is chosen when getting the median

zeman:
getMedianRunNumber() and getMedianRun() use different methods for selecting the median run when there are an even number of runs.

getMedianRuns() chooses the lower of the two middle values whereas getMedianRunNumber() chooses the upper as it doesn’t account for the 0 index on the array.

When using the "mv" param to only keep video for the median run this can lead to a mismatch between the video data that is kept and the median that is then shown on result.php and jsonResult.php

This updates getMedianRunNumber() to use the same approach as getMedianRun() and always select the lower of the two middle runs.

Please double check I’m doing the right thing here Pat!

### DIFF
--- a/www/include/TestResults.php
+++ b/www/include/TestResults.php
@@ -165,7 +165,7 @@ class TestResults {
    * @param string $metric Name of the metric to consider for selecting the median
    * @param bool $cached False if first views should be considered for selecting, true for repeat views
    * @param string $medianMode Can be set to "fastest" to consider the fastest run and not the median. Defaults to "median".
-   * @return float The run number of the median run
+   * @return float The run number of the median run (or lower of two middle values) of $metric,
    */
   public function getMedianRunNumber($metric, $cached, $medianMode = "median") {
     $values = $this->getMetricFromRuns($metric, $cached, true);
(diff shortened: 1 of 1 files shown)


www/include/TestResults.php
```diff
@@ -186,8 +186,17 @@ public function getMedianRunNumber($metric, $cached, $medianMode = "median") {
     if ($numValues == 1 || $medianMode == "fastest") {
       return $runNumbers[0];
     }
-    $medianIndex = (int)floor($numValues / 2.0);
-    return $runNumbers[$medianIndex];
+    $medianIndex = (int)floor(((float)$numValues + 1.0) / 2.0);
```
sburnicki: Add a ` - 1` and we go from a 1-based index to a 0-based index. Then we can still simply return `$runNumbers[$medianIndex]` and drop the loop. The loop below does nothing else than choosing the value by using $current as a 1-based index and comparing it with $medianIndex (also $time is misleading, as the array contains run numbers).

With `(int)floor(((float)$numValues + 1.0) / 2.0) -1`:
* $numValues = 2  →  0
* $numValues = 3  →  1
* $numValues = 4  →  1
* $numValues = 5  →  2
* ...

So it's always the lower 0-based median index.

sburnicki: Well it might also be simplified as `(int) floor(((float)$numValues - 1.0)/2.0)` to get a 0-based index.

zeman: Sure, I was just using the same code pattern as in getMedianRun() so there's consistency in the approach. Having two different ways of doing the same thing is what led to this issue. Ideally there should only be one function that gets the median run rather than two.

sburnicki: You are right, there should be only one. The new function supports for multistep data though, and the old one doesn't. Multistep needed some big changes and is not yet supported in all parts of WPT, so it wasn't possible to stick with only one method.

Nevertheless, I think using a loop instead of using a `-1` is adding confusing to the code, which needs to be different to the original anyway.

PR #746, opened 2016-10-16
Add Playform.io to CDN list

nhristov:


pmeenan:
Do you have details on the CDN POP locations?  As best as I can tell (at least from the US) the traffic is all routed to Europe.

PR #749, opened 2016-10-19
Use EC2 Instance Profile if credentials are not provided

carlnordenfelt:
Refactored EC2 and S3 implementations to use AWS environment credentials (i.e. EC2 Instance Profile) if key/secret are not provided.

Note that this PR has not been tested at all. Consider this PR to currently be a placeholder for discussion :)

I recommend adding ?w=1 to the url as there was apparently some whitespace changes done by my IDE that I didn't notice before pushing. I.e: https://github.com/WPO-Foundation/webpagetest/pull/749?w=1

### DIFF
--- a/www/archive.inc
+++ b/www/archive.inc
@@ -3,9 +3,20 @@ require_once('testStatus.inc');
 require_once('logging.inc');
 require_once('common_lib.inc');
 
+function S3_GetSdkClient() {
+    require_once('./lib/S3.php');
+    global $settings;
+
+    if (array_key_exists('archive_s3_key', $settings) && array_key_exists('archive_s3_secret', $settings)) {
+        return new S3(trim($settings['archive_s3_key']), trim($settings['archive_s3_secret']), false, trim($settings['archive_s3_server']));
+    } else {
+        return new S3(null, null, false, trim($settings['archive_s3_server']));
+    }
+}
+
 /**
 * See if the file should be skipped when archiving (usually just cache files)
-* 
+*
 * @param mixed $file
 */
 function ArchiveSkipFile($file) {
@@ -22,7 +33,7 @@ function ArchiveSkipFile($file) {
 * Archive the given test if it hasn't already been archived
 * For now this will just zip and move to a location on disk
 * but will eventually integrate with the S3 archiving
-* 
+*
 * @param mixed $id
 */
 function ArchiveTest($id, $store_result = true, $delete = false) {
@@ -49,7 +60,7 @@ function ArchiveTest($id, $store_result = true, $delete = false) {
                     SaveTestInfo($id, $testInfo);
                 }
             }
-            
+
             if (!$ret) {
               $lock = LockTest($id);
               if (isset($lock)) {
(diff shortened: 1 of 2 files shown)


www/ec2/ec2.inc.php
```diff
@@ -385,68 +406,65 @@ function EC2_DeleteOrphanedVolumes() {
...
-      if (isset($response['Regions'])) {
-        foreach ($response['Regions'] as $region)
-          $regions[] = $region['RegionName'];
-      }
-    } catch (\Aws\Ec2\Exception\Ec2Exception $e) {
-      $error = $e->getMessage();
-      EC2LogError("Listing running EC2 instances: $error");
+
+  $locations = EC2_GetAMILocations();
+  $ec2 = EC2_GetSdkClient();
```
pmeenan: Either move this inside the try/except block or (if it is handled above), check for a null $ec2 (same for all occurrences below)

www/archive.inc
```diff
@@ -3,9 +3,20 @@ require_once('testStatus.inc');
 require_once('logging.inc');
 require_once('common_lib.inc');
 
+function S3_GetSdkClient() {
+    require_once('./lib/S3.php');
```
pmeenan: I don't think the s3 library that is being used supports IAM roles.  It only takes credentials explicitly.  The s3 archiving probably needs to be moved to use the official AWS SDK (like the EC2 code does) and fall back to the s3 lib for Internet Archive storage.

www/ec2/ec2.inc.php
```diff
@@ -13,14 +13,37 @@
...
+
+    if ($awsRegion) {
+        $sdkSettings['region'] = $awsRegion;
+    }
+    if ($awsAccessKey && $awsSecretKey) {
+        EC2Log('Creating EC2 SDK Client using access key credentials');
+        $sdkSettings['key'] = $awsAccessKey;
+        $sdkSettings['secret'] = $awsSecretKey;
+    }
+    return \Aws\Ec2\Ec2Client::factory($sdkSettings);
```
carlnordenfelt: Added try/catch with logging and rethrowing the error

www/ec2/ec2.inc.php
carlnordenfelt: Moved and init function will rethrow.

www/archive.inc
carlnordenfelt: I'll take a closer look at the S3 lib and see what it does and does not support

Z3r0Sum:
@carlnordenfelt Are you going to be finishing this off?  This issue was opened in 2015 and it's equally unacceptable in 2017 :)  I could attempt to help in 2-3 weeks if this is just dead in the water?  @pmeenan if this was to be picked up now would there be a lot of re-factoring?  I haven't looked through the codebase on a critical level.

pmeenan:
It shouldn't be too bad.  The AWS interface points are only in a couple of places (mostly the scaling logic and the archive logic) and moving thouse touch points to a newer library should be pretty clean.  

The only thing that needs to be watched out for is the archiving code needs to fall back to the old library if the http archive settings are set because the Internet archive uses an s3-like API that the old code supports but the new code may not.

carlnordenfelt:
@Z3r0Sum Hi, I had honestly completely forgotten about this issue and given that this was over a year ago I don't see myself picking it up again so feel free to to do your worst with it :)

lirangabay:
@tkadlec @carlnordenfelt, is there any update to support IAM Roles instead of AWS access keys for archive_s3?

PR #769, opened 2016-12-02
adding Netlify

fool:
Our CDN is identified by at least one of a *.netlify.com hostname or a header, "server: Netlify".


fool:
I see that you asked another PR for its PoP locations.  We have about a dozen, US+Europe+South America+Asia+Australia - more details available if needed :)

Issue #695, opened 2016-08-23
JSON Results: Request data has wrong data type


sburnicki:
This is not an urgent issue, but I open it as a reminder:
In out JSON results, most of the request data has the wrong data type and is string instead of numeric.

I guess the actual problem occurs in `LoadRequests()` in `object_detail.inc` where the values are loaded from a text file, but not being converted to an appropriate numeric type.

PR #794, opened 2017-01-22
WIP: Store sizes and times as numbers, not strings

tobli:
This is a first attempt to address #695. Some extended fields in the HAR (fields beginning with _) are represented as strings, even though they are numeric. I don't have much experience in either PHP or the WPT code base, so I'm posting this in rough form to get some initial feedback. Am I on the right track? Is it a worthwhile patch that could get accepted if polished?

The patch is based on a number of observations/assumptions:
- fields representing sizes (bytes) and times (ms) are numeric at the source (see [results.cc#L1305](https://github.com/WPO-Foundation/webpagetest/blob/cdd044d530edf147f96a32d1fac0dec63f6d2e7f/agent/wpthook/results.cc#L1305)), but as written to an intermediate text file and parsed as strings.
- some fields as conceptually booleans (e.g. _was_pushed) but are stored as 0 or 1.
- for page level extended fields (many with the same name as for assets) there's an explicit cast to int where applicable (see [page_data.inc#L367](https://github.com/WPO-Foundation/webpagetest/blob/master/www/page_data.inc#L367)).
- implicit conversion in PHP from string to number takes place when strings are used in numeric operations, leading to surprising results such as
```
                "_ttfb_start": "494",
                "_ttfb_end": 831,
```
- the unit tests (e.g. in [objectDetailTest.php#L30](https://github.com/WPO-Foundation/webpagetest/blob/master/www/tests/objectDetailTest.php#L30)) does implicit conversion between string and number, so it doesn't seem to matter if the expected result for a field is 42 and the generated result is '42', further masking the issue.


www/object_detail.inc
```diff
@@ -239,8 +291,17 @@ function LoadRequests($request_file_name) {
...
+                    if (array_key_exists($i + 1, $needed_columns) && strlen($trimmed_value)) {
+                        $request_key = $needed_columns[$i + 1];
+                        if (array_key_exists($request_key, $int_fields)) {
+                          $request[$request_key] = (int)$trimmed_value;
+                        } else if (array_key_exists($request_key, $float_fields)) {
+                          $request[$request_key] = (float)$trimmed_value;
+                        } else {
+                          $request[$request_key] = $trimmed_value;
+                        }
+                    }
```
pmeenan: Rather than a lookup table I'd rather use a runtime check to see if it is numeric like [here](http://php.net/manual/en/function.is-numeric.php#107326).

Basically turn it into this:
```php
$trimmed_value = trim($value);
if (isset($needed_columns[$i + 1]) && strlen($trimmed_value)) {
  if (is_numeric($trimmed_value)) {
    $request[$needed_columns[$i + 1]] = $trimmed_value + 0;
  } else {
    $request[$needed_columns[$i + 1]] = $trimmed_value;
  }
}
```

tobli: 👍 I'll fix and will push an update

PR #795, opened 2017-01-23
Added Naver to the list of known CDNs

difro:
Please add [Naver](https://en.wikipedia.org/wiki/Naver)'s CDN to the known CDN list.

```
$ $ curl -si -D- -o /dev/null 'https://ssl.pstatic.net/sstatic/search/favicon/favicon_140327.ico'
HTTP/1.1 200 OK
Date: Tue, 03 Jan 2017 12:41:55 GMT
Server: Testa/4.8.6
Last-Modified: Thu, 27 Mar 2014 06:34:55 GMT
ETag: "1536-4f590c88381c0"
Accept-Ranges: bytes
Content-Length: 5430
Cache-Control: max-age=31536000
Expires: Wed, 03 Jan 2018 12:41:55 GMT
Vary: Accept-Encoding
Content-Type: image/x-icon
Age: 1709048
Strict-Transport-Security: max-age=31536000
```

### DIFF
--- a/agent/wpthook/cdn.h
+++ b/agent/wpthook/cdn.h
@@ -219,7 +219,8 @@ CDN_PROVIDER_HEADER cdnHeaderList[] = {
   {"server", "Sucuri/Cloudproxy", "Sucuri Firewall"},
   {"x-sucuri-id", "", "Sucuri Firewall"},
   {"server", "Netlify", "Netlify"},
-  {"section-io-id", "", "section.io"}
+  {"section-io-id", "", "section.io"},
+  {"server", "Testa/", "Naver"}
 };
 
 // Specific providers that require multiple headers


pmeenan:
Does Naver run their own CDN or is that Naver's static domain which runs on top of other CDN's (or a hybrid)?  Testing from Virginia, it looks like ssl.pstatic.net is hosted on Akamai.

difro:
Yeah, it's a hybrid.

Only the requests originating from S.Korea are directed to our own CDN. The rest of the world sees the other CDN's.

pmeenan:
It will take a few days to actually show up.  Changes only take effect when I roll out a new agent and I usually don't roll just for CDN changes (but it won't be past the end of the week for sure).

PR #808, opened 2017-02-09
Create .dockerignore

jangrewe:
Don't send the `.git` folder to the Docker context when building the image, which saves (currently) 627MB during the build.


sburnicki:
Good point. Actually the file could also look like
```
*
!Dockerfile
!www
```
Since we only need the www folder (don't know about the Dockerfile itself). This effectively reduces the build context to 56MB :)

jangrewe:
The Dockerfile is ignored automatically, IIRC, but `www` is actually required - else you'd have no content to display!

sburnicki:
Okay, then it could come down to
```
*
!www
```
which means: Exclude everything but the `www` folder (see https://docs.docker.com/engine/reference/builder/#/dockerignore-file)

jangrewe:
Well, you'd also have to exclude the `docker` folder, because that is also partially copied. I think we're fine by at least excluding `.git` withouth breaking anything.

sburnicki:
Okay ;)

PR #813, opened 2017-02-13
feat: add lighthouse option to task

patrickhulce:


www/runtest.php
```diff
@@ -1311,6 +1313,17 @@ function ValidateParameters(&$test, $locations, &$error, $destination_url = null
 
             if( !$test['aftEarlyCutoff'] && $settings['aftEarlyCutoff'] )
                 $test['aftEarlyCutoff'] = $settings['aftEarlyCutoff'];
+
+            if ( $test['lighthouse'] ) {
+              $lh_error = '';
+              if ( strpos($test['browser'], 'Chrome') === FALSE )
+                $lh_error = 'Lighthouse only works on Chrome';
+              if ( strlen($test['script']) )
```
patrickhulce: This was the only setting I immediately see will make the code paths completely divergent, everything else can likely be ignored, but can make them errors if desired.

www/runtest.php
```diff
@@ -1311,6 +1313,17 @@ function ValidateParameters(&$test, $locations, &$error, $destination_url = null
 
             if( !$test['aftEarlyCutoff'] && $settings['aftEarlyCutoff'] )
                 $test['aftEarlyCutoff'] = $settings['aftEarlyCutoff'];
+
+            if ( $test['lighthouse'] ) {
+              $lh_error = '';
+              if ( strpos($test['browser'], 'Chrome') === FALSE )
```
patrickhulce: Didn't see any setting that reveals if the agent is node or not, anything I missed?

www/runtest.php
pmeenan: That should pick it up but it might be easier to just flag it in the UI as "Chrome Mobile devices only" or something and just let lighthouse fail to run silently otherwise.  When it moves to support desktop it would include "Canary" as well but people can also rename the browser labels to be anything.

www/runtest.php
pmeenan: No, the server doesn't know anything about how the agents run the tests.

PR #814, opened 2017-02-15
feat: add lighthouse option to node agent

patrickhulce:


agent/js/src/wd_server.js
```diff
@@ -155,7 +155,8 @@ WebDriverServer.prototype.initIpc = function() {
  */
 WebDriverServer.prototype.init = function(args) {
   'use strict';
-  args.startURL = args.task['Capture Video'] ? ORANGE_PAGE_URL_ : BLANK_PAGE_URL_;
+  args.startURL = args.task['Capture Video'] && !args.task.lighthouse ? 
+      ORANGE_PAGE_URL_ : BLANK_PAGE_URL_;
```
patrickhulce: I think it'll be better to have lighthouse change it's about:blank to arbitrary this rather than try to orchestrate the tab ID navigation, etc

agent/js/src/wd_server.js
```diff
@@ -1142,8 +1160,11 @@ WebDriverServer.prototype.runPageLoad_ = function(browserCaps) {
     this.startChrome_(browserCaps);
```
patrickhulce: this doesn't seem to be a problem in practice even though I think it should... Does Chrome always get closed in between runs?

agent/js/src/wd_server.js
```diff
@@ -896,12 +896,20 @@ WebDriverServer.prototype.clearPageAndStartVideoDevTools_ = function() {
       this.networkCommand_('setExtraHTTPHeaders', {'headers': this.task_.headers});
     this.pageCommand_('enable');
   }
+};
+
+/**
+ * Blanks out the browser at the beginning of a test, using DevTools.
+ * @private
+ */
+WebDriverServer.prototype.holdBlankAndStartVideoDevTools_ = function() {
```
patrickhulce: Other name ideas welcome, just wanted to update from what I assume was drift over time.

agent/js/src/wd_server.js
pmeenan: In the case of video capture, the orange page needs to be up at the time of the start of the capture.  Unless lighthouse is going to run the screenrecord, the page needs to be loaded before lighthouse starts.

agent/js/src/wd_server.js
pmeenan: sgtm

agent/js/src/wd_server.js
```diff
@@ -896,12 +896,20 @@ WebDriverServer.prototype.clearPageAndStartVideoDevTools_ = function() {
...
+ */
+WebDriverServer.prototype.holdBlankAndStartVideoDevTools_ = function() {
+  'use strict';
   if (1 === this.task_['Capture Video']) {  // Emit video sync, start recording
     this.getCapabilities_().then(function(caps) {
       if (!caps.videoRecording) {
         return;
       }
-      if (!this.browser_.isBlackBox) {
+      if (!this.browser_.isBlackBox && !this.task_.lighthouse) {
```
pmeenan: Actually, the getResourceTree bit is no longer used and I can clean it up.  The main thing is that you still need to start video capture and wait the 500ms in lighthouse mode (and only skip the tracing).  It might be cleaner to just have the lighthouse check inside of scheduleStartTracingIfRequested_()

agent/js/src/wd_server.js
patrickhulce: Discussed separately, but conclusion is that this will not be needed for now. 

I will investigate potential IPC communication in lighthouse to notify the agent that the orange screen is being displayed and will fallback to trimming the video up until lighthouse's orange page rather than WPTs.

patrickhulce:
@pmeenan did you want to wait for your cleanup to merge or do I have action items first? Since the recording details involve adding LH implementation to WPT anyway I was hoping to save that for a future PR and keep the reviews small and incremental.

patrickhulce:
bump :) any action required on my part for this?

pmeenan:
Doesn't look like the PR was updated for the comments above.  Specifically, keeping the orange page and waiting the 500ms before starting lighthouse.

patrickhulce:
Ah, those were my clarifying comments, I had thought we settled on addressing those later hence the "this will not be needed for now". Will do.

pmeenan:
np, I can go either way since it doesn't affect the default path.  If you are ok with how it is now and cleanup coming later as integration progresses I'm happy to merge.

patrickhulce:
Yep I've already done it locally and will be in next PR coming imminently! :)

PR #825, opened 2017-03-02
feat: add basic lighthouse functionality

patrickhulce:
cc: @paulirish @ebidel @brendankenny

**The Lighthouse**

**The Run**
image

**The Report** (video is available on mobile)
image

### DIFF
--- a/agent/js/src/wd_server.js
+++ b/agent/js/src/wd_server.js
@@ -1028,13 +1028,34 @@ WebDriverServer.prototype.onTracingMessage_ = function(message) {
 };
 
 /**
- * Runs lighthouse with the options specified by the task and dumps data into 
+ * Runs lighthouse with the options specified by the task and dumps data into
  * a directory.
  * @private
  */
 WebDriverServer.prototype.runLighthouse_ = function() {
-  // TODO(phulce): actually run lighthouse
-  this.onPageLoad_(new Error('Lighthouse not yet implemented!'));
+  this.lighthouseFile_ = path.join(this.runTempDir_, 'lighthouse.html');
+  this.app_.timeout(1000, 'Waiting for Chrome to be available...');
+  process_utils.scheduleExec(this.app_,
+    'lighthouse',
+    [
+      '--disable-device-emulation',
+      '--disable-network-throttling',
+      '--save-assets',
+      '--port', this.browser_.devToolsPort_,
+      '--output', 'html',
+      '--output-path', this.lighthouseFile_,
+      this.task_.url,
+    ], undefined, this.timeout_).then(function () {
+    var devtoolsLogFile = path.join(this.runTempDir_, 'lighthouse-0.devtoolslog.json');
+    this.devToolsMessages_ = JSON.parse(fs.readFileSync(devtoolsLogFile, 'utf8'));
+
+    this.traceFile_ = path.join(this.runTempDir_, 'lighthouse-0.trace.json');
+    var contents = JSON.parse(fs.readFileSync(this.traceFile_, 'utf8'));
+    fs.writeFileSync(this.traceFile_, JSON.stringify(contents));
+    this.onPageLoad_();
+  }.bind(this), function (err) {
+    this.onPageLoad_(err);
+  }.bind(this));
 };
 
 /**
@@ -1590,7 +1611,7 @@ WebDriverServer.prototype.done_ = function() {
 
     this.scheduleNoFault_('Capture Screen Shot', function() {
       if (!this.task_.lighthouse)
-        this.takeScreenshot_('screen', 'end of run');
+        return this.takeScreenshot_('screen', 'end of run');
     }.bind(this));
     if (this.videoFile_) {
       this.scheduleNoFault_('Stop video recording',
@@ -1668,6 +1689,7 @@ WebDriverServer.prototype.done_ = function() {
           devToolsFile: devToolsFile,
           screenshots: this.screenshots_,
           traceFile: this.traceFile_,
+          lighthouseFile: this.lighthouseFile_,
           userTimingFile: this.userTimingFile_,
           cpuSlicesFile: this.cpuSlicesFile_,
           scriptTimingFile: this.scriptTimingFile_,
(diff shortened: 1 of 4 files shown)


agent/js/src/wd_server.js
```diff
@@ -1028,13 +1028,34 @@ WebDriverServer.prototype.onTracingMessage_ = function(message) {
...
  * a directory.
  * @private
  */
 WebDriverServer.prototype.runLighthouse_ = function() {
-  // TODO(phulce): actually run lighthouse
-  this.onPageLoad_(new Error('Lighthouse not yet implemented!'));
+  this.lighthouseFile_ = path.join(this.runTempDir_, 'lighthouse.html');
+  this.app_.timeout(1000, 'Waiting for Chrome to be available...');
+  process_utils.scheduleExec(this.app_,
+    'lighthouse',
```
paulirish: Probably up to Pat, but should we put the lighthouse install instructions in a readme?

Presumably it's `npm install -g GoogleChrome/lighthouse` + confirm npm globals are in your PATH?

patrickhulce: Yeah the most helpful instructions I actually found were over in the [google sites page](https://sites.google.com/a/webpagetest.org/docs/private-instances/node-js-agent/setup) rather than any README so I figured we'd add it there, but even better to find a place here if there's one I missed!

agent/js/src/wd_server.js
```diff
@@ -1028,13 +1028,34 @@ WebDriverServer.prototype.onTracingMessage_ = function(message) {
...
+    [
+      '--disable-device-emulation',
+      '--disable-network-throttling',
+      '--save-assets',
+      '--port', this.browser_.devToolsPort_,
+      '--output', 'html',
+      '--output-path', this.lighthouseFile_,
+      this.task_.url,
+    ], undefined, this.timeout_).then(function () {
+    var devtoolsLogFile = path.join(this.runTempDir_, 'lighthouse-0.devtoolslog.json');
```
ebidel: is this indentation intentional?

www/lighthouse.php
```diff
@@ -0,0 +1,26 @@
...
+  $ok = false;
+  if (isset($testPath) && is_dir($testPath)) {
+    $file = "${run}_lighthouse.html.gz";
+    $filePath = "$testPath/$file";
+    if (is_file($filePath)) {
+      $ok = true;
+
+      // Cache for a year
+      header('Last-Modified: ' . gmdate('r'));
+      header('Expires: '.gmdate('r', time() + 31536000));
```
ebidel: space around `.`

ebidel: IIRC `Expires` is old and everyone supports `Cache-Control` now. Do we need it?

https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching?hl=en#cache-control

ebidel:
Does our snazy LH animation come into the mix anywhere, or is that just for PR description fun?

agent/js/src/wd_server.js
pmeenan: How big is it and is it feasible to include directly?  Worst case I'll add logic to the agent at startup to look for it and try to install if it isn't there (and automatically update if it is) but if we can eliminate the need for manually installing then all of the existing agents will pick up lighthouse support automatically.

Also, does running it as "lighthouse" work on Windows as well or does Node need to explicitly be referenced?  Just checking since the Node agent runs across all platforms.

www/lighthouse.php
pmeenan: The WPT code is old so it's just left over :-)  It's fine to remove it but it also doesn't hurt since cache-control takes precedence.

www/result.inc
```diff
@@ -125,8 +125,15 @@ $page_description = "Website performance test result$testLabel.";
...
                 </div>
                 <?php
+                    $fvMedian = $testResults->getMedianRunNumber($median_metric, false);
+                    $rvMedian = $testResults->getMedianRunNumber($median_metric, true);
+
                     // create a friendlier (unique) name for the fi
                     $fileUrl = GetFileUrl($url);
+                    if ( $test['testinfo']['lighthouse'] && $fvMedian )
+                    {
+                      echo "<a href=\"/lighthouse.php?test=$id&run=$fvMedian\">Lighthouse report</a><br>";
```
pmeenan: This works for now but I'd really like to give it much more prominent treatment.  The grades section in the header is a perfect place since the grades won't be calculated in the lighthouse case.  If the score can be pulled out of the HTML then the UI can have a large:

Lighthouse Score: 85/100 (click for full report)

And maybe also the logo ;-)

ebidel:
+1 on iterating on the UI and figuring out more prominent bits as we go.
Including the logo will help users know this is the same tool they've used
elsewhere.

www/lighthouse.php
patrickhulce: I'll remove

agent/js/src/wd_server.js
patrickhulce: this line yes :) though the above should be indented 2 more spaces. it's used elsewhere like this, will fix

agent/js/src/wd_server.js
patrickhulce: IMO the maintenance burden of keeping it as a submodule or checking in the result of npm install here is more trouble than letting `npm install -g` handle it and adding an auto-update check, but it's a bit safer so I'm open to either.

I'm not on Windows anywhere with node, so I can't verify it, but we've recently had a contributor making sure our build runs on appveyor which I believe removed our *nix specific hashbangs and such. If your path is setup correctly it should, but I can validate on my machine at home tonight.

patrickhulce:
> Does our snazy LH animation come into the mix anywhere, or is that just for PR description fun?

Just fun, but it will! That GIF was from my local  custom LH-WPT launch page ;)

I also vote to iterate on the UI, adding the score, etc.

ebidel:
Woot! cc @igrigorik

igrigorik:
\o/ ... woot!

Pat: when might we see this live on webpagetest.org? :)

pmeenan:
Probably next week.  I want to do some testing first and then since I have ~60 mobile agents I really want to teach the agent code to install and update lighthouse automatically so I don't have to manually configure all of them.

pmeenan:
There go my hopes for an easy install :(  Apparently the Node version on the Pi's is 4.x (which is what is in the repo for jessie) so I need to go through and upgrade them all to 7.x anyway.  Starting on the testing now and I still plan to have it up and running next week.

PR #826, opened 2017-03-02
Add visualComplete85 metric

JeroenVdb:
I really like the visualComplete85 that Lighthouse (I think?) introduced.

I already use the visualComplete metric but this one spikes when a pages has a video playing ad. Or sometimes stays 99% for a very long time for a very small change.

This visualComplete85 can mitigate these small changes from metric dashboards.

I don't know where the 85% number originally comes from but for now lets keep it in line with other tools?

### DIFF
--- a/www/video/visualProgress.inc.php
+++ b/www/video/visualProgress.inc.php
@@ -157,6 +157,8 @@ function GetVisualProgressForStep($localPaths, $runCompleted, $options = null, $
     foreach($frames['frames'] as $time => &$frame) {
       if ($frame['progress'] > 0 && !array_key_exists('startRender', $frames))
         $frames['startRender'] = $time;
+      if (!$frames['visualComplete85'] && $frame['progress'] >= 85)
+        $frames['visualComplete85'] = $time;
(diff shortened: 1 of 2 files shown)


www/video/visualProgress.inc.php
```diff
@@ -157,6 +157,8 @@ function GetVisualProgressForStep($localPaths, $runCompleted, $options = null, $
     foreach($frames['frames'] as $time => &$frame) {
       if ($frame['progress'] > 0 && !array_key_exists('startRender', $frames))
         $frames['startRender'] = $time;
+      if (!$frames['visualComplete85'] && $frame['progress'] >= 85)
```
pmeenan: Since visualComplete85 isn't initialized, can you change this to:
```
if (!isset($frames['visualComplete85']) && $frame['progress'] >= 85)
```
(or initialize it to zero like visualComplete above but I like the idea that it won't be set if it wasn't calculated)

JeroenVdb:
@pmeenan fixed the check for visualComplete85.

Issue #811, opened 2017-02-09
Firefox addon needs to be converted to WebExtension for 57


digitarald:
I can reach out to the Firefox addon team. Is there anything that web extensions can not do that WPT needs?

cpeterso:
Firefox 57 (November 2017) will drop support for legacy add-ons and only support [WebExtensions](https://developer.mozilla.org/en-US/Add-ons/WebExtensions). Unfortunately, Firefox's WebExtension API does not yet include all of the Chrome extension APIs used by the wptdriver. Some missing APIs:

* chrome.management.getAll ([Firefox bug 1282981](https://bugzilla.mozilla.org/show_bug.cgi?id=1282981))
* chrome.management.launchApp
* contentSettings
* debugger
* some differences in tabs.create

pmeenan:
FWIW, it doesn't need to be the way things are right now.  The main needed features are:
- Navigate a tab
- Get onload and error notifications for the navigation
- Run JS as if it was running from the dev console

As gravy, getting the request/response timings, headers and bodies and being able to do it all remotely would eliminate the need for an extension at all.

luser:
Just curious--is there a reason webpagetest doesn't use WebDriver or something similar? Chrome and Firefox both have good built-in support for remote protocols based on that.

pmeenan:
That works fine for driving the browser (to some extent) but not for extracting the timing of all of the resources, the headers, response bodies, etc.

Granted, for the current desktop implementation that uses API hooking to measure that stuff, webdriver may be a viable option to replace the Firefox extension.

PR #855, opened 2017-04-03
Ported Firefox extension to WebExtension

squarewave:
Brought over a large amount of the Chrome extension code, and as such found it simplest to continue compiling via closure. A few tricky bits to highlight:

- `exec` is now accomplished via posting a message to a script which is loaded into the page. This allows us to remove access to WebExtension APIs.
- `collectstats` is also accomplished via the same mechanism, to ensure we get access to the page's window.performance object.

Something that came up in [our bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1348394#c7) tracking this, was whether we should consolidate this code and the Chrome extension code and just wrap the things that the Chrome extension uses (mostly the debugger API) in conditionals. I don't think this would be too crazy of a task, but I figured I should ask here.


squarewave:
Also, I was wondering what the best strategy for testing this is. So far, I've done the following:

- Run all the tests in allTests.js
- Run all the fake commands in firefox/extension/chrome/content/fakeCommandSource.js
- Stand up a local instance of WebpageTest and run it against https://facebook.com with the WebExtension in the profile, then compare the results to the same test with the old Firefox extension.

pmeenan:
FWIW, I'm planning on porting the Firefox testing over to the [new agent](https://github.com/WPO-Foundation/wptagent) (in the next few weeks).

Some combination of webdriver to automate/exec and extension or remote API for getting the request timings, headers and bodies.

As far as testing the current agent code, I don't think allTests.js has been updated in a long time.  Usually the testing is just manual testing (like you have done).  Also, appologies for closure - it was done a long time ago and overly complicates the extension.

About the posting for exec, are there any sandbox issues to be aware of when interacting with the page?

squarewave:
> About the posting for exec, are there any sandbox issues to be aware of when interacting with the page?

What kinds of issues did you have in mind? It should be roughly equivalent to running JS in the console, but if you have some specifics you're concerned about I can ask around.

pmeenan:
Most of the issues revolved around interacting with the DOM or page variables but it's currently in kind of a rough state so hopefully it isn't too bad.  I'll give it a try later this afternoon and see how things look.

cpeterso:
@squarewave Can we close WPT issue #811 as fixed by this PR?

PR #858, opened 2017-04-07
Strip CSP header to allow exec script injection

### LINKED ISSUE #857: Problems with "exec" on Firefox since new extension (371)
Since using the new extension (#855), quite some scripts stopped working with Firefox.
For example:
```
navigate https://login.audi.com/login/#/de/DE/login
sleep 3
execAndWait document.querySelector(".login_loginform_registernow").click()
```
This script usually results in two steps, the second after clicking on a link.
This works with [Chrome](http://www.webpagetest.org/result/170407_T0_KKS/) and also worked with Firefox before. Now, Firefox [doesn't execute](http://www.webpagetest.org/result/170407_M4_KJP/) the JavaScript anymore.

I also tried different JavaScript commands (like removing DOM nodes, setting input, etc), they all do not work. Interestingly only specific websites are affected, like the one refrenced. On others I don't have this problem.

When I watched Firefox on an gent running the code I saw for example
```
InPageCommandRunner got a command:  Object { command: "evalInPage", target: "document.querySelector(".lo…" }
```
in the console, but nothing happened. When I pasted the same "target" into the console, it was correctly executed.

Any hints on why this behavior might happen or how I can track this down?

squarewave:
Closes #857 

Need review on this from a security perspective. Are we okay with disabling content security policies for machines running the WPT agent? It definitely fixes the issue for the audi site mentioned in the issue.


pmeenan:
Without a better option I think we'd have to.  Do you happen to know if there is much overhead in a blocking response hook (I know Chrome's implementation has crazy overhead at least on the request side)?

Also, any chance I can convince you to also update the signed xpi?  That would save me a step :-)

squarewave:
> Do you happen to know if there is much overhead in a blocking response hook (I know Chrome's implementation has crazy overhead at least on the request side)?

Hmm, I don't unfortunately.

> Also, any chance I can convince you to also update the signed xpi? That would save me a step :-)

Of course!

PR #864, opened 2017-04-24
Adjust TASK_INTERVAL to 500 from 5000

### LINKED ISSUE #863: Firefox idles a lot of localhost:8000, stretching out test runs
Not sure if this happened before the #855 , but Firefox waits a long time on the initial local page before loading the actual test page. In Chrome as comparison the local page is barely visible and immediately triggers the next page load.

squarewave:
Closes  #863

This was causing tests to run longer than they should.


pmeenan:
Thanks.  I'll roll a new agent update with it later this afternoon.

PR #885, opened 2017-06-07
Change all hardcoded paths to values from settings.ini

olegfedoseev:
I wanted the ability to move all writeable directories to one common prefix. For example to use with one VOLUME in Dockerfile or to move them to different hard drive.

Changes quite extensive, but most of the time straightforward.
I wasn't able to verify all of them because I don't use and, honestly, don't fully understand half of the functionality :)

Hope it'll be useful not only for me.

### DIFF
--- a/www/aggregate.php
+++ b/www/aggregate.php
@@ -126,7 +126,7 @@ if( isset($tests) )
             GetTestStatus($test['id']);
             $label = $test['l'];
             $url = $test['u'];
-            $testPath = './' . GetTestPath($test['id']);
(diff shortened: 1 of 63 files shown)


olegfedoseev:
Looks like it isn't so simple as I thought before. I will keep working on it.
At the moment it shouldn't be merged.

olegfedoseev:
Fix all broken URLs I can find. I have this branch as my private instance and at least all the main functionality works fine.
I rebased on latest master, and looking forward to feedback!

PR #908, opened 2017-07-21
fix incorrect assignment in 58ec2814

xeonchen:
the assignment was incorrectly typed as comparison.


sburnicki:
Great catch, thanks! We experienced problems while polling for XML results of slow/long tests. I already investigated the issue for some hours, but didn't find the problem. Now this should fix it :)

Issue #914, opened 2017-08-09
Vector logo & icon


sergeychernyshev:
Tried to find a vector version of WPT logo or icon and had no luck.

I asked my son to vectorize one so if there is none exists from the source, we can contribute it.

sergeychernyshev:
They seem to have a different font (serif instead of sanserif with 3different thicknesses like on current logo)

PR #916, opened 2017-08-11
Updated SVG logo and icon

sergeychernyshev:
Adds new versions of logo and separate files for icons (issue #914).


pmeenan:
Thanks.  I think the original relied on specific, custom fonts to be installed so this is much cleaner.

PR #921, opened 2017-08-23
Explicitly convert domain name to string to avoid problems in edge cases

sburnicki:
In a test I got an edge case where a request to `http://36170252` was captured (I have no idea why or what this actually is). In the domain breakdown, the domain name was treated as integer instead of string which broke the whole chart.

Explicitly converting the value to String should fix these edge cases.

### DIFF
--- a/www/domains.php
+++ b/www/domains.php
@@ -186,12 +186,12 @@ if (array_key_exists('f', $_REQUEST) && $_REQUEST['f'] == 'json') {
             bytes.addColumn('number', 'Bytes');
             bytes.addRows(numData);
             for (var i = 0; i < numData; i++) {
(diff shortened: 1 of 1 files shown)


pmeenan:
I've never seen conversion in php done like that and couldn't find any docs.  Does it work as expected?  As best as I can tell the [recommended ways](http://php.net/manual/en/language.types.string.php#language.types.string.casting) are either to cast to (string) or to use the strval() function.  The only case I had seen with a string constructor was for passing the length in explicitly.

sburnicki:
It's the JavaScript part of the file. It's using [this method](https://www.w3schools.com/Jsref/jsref_string.asp).

pmeenan:
Doh, sorry - thanks.

PR #927, opened 2017-08-25
Adding flag to xperf provider to reduce ETL size

vibaldem:
This flag should tell xperf to only record the events that are being parsed and read by the script.  It will reduce the ETL size and improve overall execution time for the Edge agent.


pmeenan:
Thanks.  btw, I ported the Edge support to [wptagent](https://github.com/WPO-Foundation/wptagent) (the new cross-platform agent) so things like multi-step scripts should work now.

Still have a bunch of features to add that will probably require an extension but it should already be more capable that what wptdriver can do.

PR #948, opened 2017-10-06
add Video as a MimeType to waterfall and content breakdown

dougsillars:
Added file count and bytecount for Video mimetyes (I used a similar profiling as in the HTTPArchive count).  Added a blue/teal colour to the waterfalls and the content breakdown:
image

Note that the Video colour is VERY similar to the Flash colour.
image
image

This may not be as important as flash usage is dropping as video usage is increasing.

Doug


pmeenan:
LGTM.  Given Flash's deprecation schedule in most browsers we can probably remove it soon and just let it fall into other (at least for the visualizations, probably good to continue to report it separately).

PR #957, opened 2017-10-24
New setting: EC2.IdleTerminateMinutes

josephscott:
This setting is for Linux EC2 instances, which are now billed by the second,
with a minimum of 60 seconds.  The name says it all, it is an int that provides
the number of minutes to let an instance sit idle before terminating it.

I have included comments around this setting, because some caution is needed.
If you push the number too low the agent will not have time to spin up and
ask for tests before being considered idle.

Windows instances are still billed per hour.  When EC2.IdleTerminateMinutes
is not set then the existing termination logic is still applied.

https://aws.amazon.com/about-aws/whats-new/2017/10/announcing-amazon-ec2-per-second-billing/

This has been working fine for the Linux agents that I've been testing.  I have not tested for Windows agents, but there should be no change there ( assuming EC2.IdleTerminateMinutes is not set ).

### DIFF
--- a/www/ec2/ec2.inc.php
+++ b/www/ec2/ec2.inc.php
@@ -124,8 +124,16 @@ function EC2_StartInstance($ami) {
 
 /**
 * Terminate any EC2 Instances that are configured for auto-scaling
-* if they have not had work in the last 15 minutes and are close
-* to an hourly increment of running (since EC2 bills hourly)
+* if they have not had work.
+*
+* Linux agents are billed per second ( with a minimum of 60 seconds ), and
+* are terminated after being idle for EC2.IdleTerminateMinutes, from the
+* settings.
+*
+* Windows agents are billed hourly and terminate if they have not had any
+* work in the last 15 minutes and are close to an hourly increment of running.
+*
+* https://aws.amazon.com/about-aws/whats-new/2017/10/announcing-amazon-ec2-per-second-billing/
 * 
 */
 function EC2_TerminateIdleInstances() {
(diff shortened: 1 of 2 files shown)


pmeenan:
If I remember right, it should be safe on the startup and not consider the agents idle until after they have connected (though you really don't want to churn too tightly and have all of your instance time go to starting up).

www/ec2/ec2.inc.php
```diff
@@ -158,9 +166,16 @@ function EC2_TerminateIdleInstances() {
       }
     }
 
+	$idleTerminateMinutes = GetSetting("EC2.IdleTerminateMinutes");
+	$minutes = $instance['runningTime'] / 60.0;
+	if ($idleTerminateMinutes) {
+		$timeCheck = ( $minutes > $idleTerminateMinutes );
```
pmeenan: This just checks that it has been running for longer than IdleTerminateMinutes, not that it has been idle that long.  The actual idle check is below where $terminate=false is set.  It is still hard-coded to not terminate if it got work in the last 15 minutes.

If that's intended then the setting is probably better to be called "EC2.CheckTerminateMinutes"

josephscott: Ah, I see what you mean.  I missed that clause further down.  I missed this because I was doing a tight test with a single instance, so the real idle check wasn't always being applied.

So something like this:

```
diff --git a/www/ec2/ec2.inc.php b/www/ec2/ec2.inc.php
index 6ac24f77..69a206a8 100644
--- a/www/ec2/ec2.inc.php
+++ b/www/ec2/ec2.inc.php
@@ -194,9 +194,9 @@ function EC2_TerminateIdleInstances() {
           }
         }
         
-        // Keep the instance if the location had work in the last 15 minutes
-        // and if this instance has checked in recently
-        if (isset($lastWork) && isset($lastCheck) && $lastWork < 15 && $lastCheck < 15)
+        // Keep the instance if the location had work in the last^M
+        // EC2.IdleTerminateMinutes and if this instance has checked in recently^M
+        if (isset($lastWork) && isset($lastCheck) && $lastWork < $idleTerminateMinutes && $lastCheck < $idleTerminateMinutes)^M
           $terminate = false;
         
         if ($terminate) {
```

Does that look reasonable?

pmeenan: That should work.  Might still not be a bad idea to have a separate setting for the minimum instance run time.

pmeenan: Sorry, one more issue.  The check needs to be inside of the foreach loop because $minutes is derived from the instance data.  The legacy path is changed otherwise (broken).

josephscott:
Yes, I should have caught that.  Updated.

PR #959, opened 2017-10-25
change to file for reduce code 

jirawat050:


pmeenan:
Any particular reason for the PR?  i.e. what does factoring the code in import.php out buy going forward?

jirawat001:
I think to write only file see Untidy

PR #960, opened 2017-10-25
add code to check one sapiku

alhudaghifari:


pmeenan:
None of this code makes any sense.  What are you trying to do?

PR #968, opened 2017-10-28
Update css 

keangkai:
use external css for separate work between css and html
(if wrong I'm sorry)


pmeenan:
The external css is not a good idea for a small number of styles.

PR #969, opened 2017-10-28
Update CDN's list

Glennmen:
Added cloudflare CDN server name: cloudflare-nginx


pmeenan:
The header match is a substring match so the server: cloudflare check right above the patch should already catch it.

Glennmen:
@pmeenan Guess you are right, after further investigating my webpagetest.org report I noticed something strange.
When I load my website from a browser it goes through Cloudflare and the response has the correct headers. 
But on the webpagetest.org there are no cloudflare headers shown in the static files response info.

PR #970, opened 2017-10-28
Update site.js

jirawat050:
edit format code for edit code


pmeenan:
I don't take reformat-only patches (and additionally, you expanded the minified jquery).  If reformatting is done as part of a functional change then it may be acceptable but not on it's own.

jirawat050:
Okay,thank you

PR #978, opened 2017-11-02
update

DarkSoulCool:


pmeenan:
Python doesn't use fi at the end of if blocks, that's just used in shell scripts (which these are not)

PR #981, opened 2017-11-07
add show text index.php to easy mode before not show link

jirawat050:
before 
screen shot 2017-11-07 at 1 11 29 pm
new 
screen shot 2017-11-07 at 1 46 25 pm

### DIFF
--- a/www/index.php
+++ b/www/index.php
@@ -140,9 +140,9 @@ $loc = ParseLocations($locations);
             <div id="test_box-container">
                 <ul class="ui-tabs-nav">
                     <li class="analytical_review ui-state-default ui-corner-top ui-tabs-selected ui-state-active"><a href="#">Advanced Testing</a></li>
-                    <li class="easy_mode"><a href=\"/easy.php\">Simple Testing</a></li>
-                    <li class="visual_comparison"><a href="/video/">Visual Comparison</a></li>
-                    <li class="traceroute"><a href="/traceroute.php">Traceroute</a></li>
+                    <li class="easy_mode"><a href="easy.php">Simple Testing</a></li>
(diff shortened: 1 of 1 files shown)


www/index.php
```diff
@@ -140,9 +140,9 @@
             <div id="test_box-container">
                 <ul class="ui-tabs-nav">
                     <li class="analytical_review ui-state-default ui-corner-top ui-tabs-selected ui-state-active"><a href="#">Advanced Testing</a></li>
-                    <li class="easy_mode"><a href=\"/easy.php\">Simple Testing</a></li>
-                    <li class="visual_comparison"><a href="/video/">Visual Comparison</a></li>
-                    <li class="traceroute"><a href="/traceroute.php">Traceroute</a></li>
+                    <li class="easy_mode"><a href="easy.php">Simple Testing</a></li>
```
pmeenan: WPT doesn't support relative paths.  You need to keep the absolute paths in place.

jirawat050: if correct edit to a href ="/easymode" ?

jirawat050: if is echo not show text

pmeenan: no, it is "/easy.php" or "/easy".  There is no /easymode rewrite rule

jirawat050:
Okai I edit finish thankyou

www/index.php
jirawat050: ok i edit finish thank you

jirawat050: Can PR complete the merge.

jirawat050:
@pmeenan  why edit correct?if not correct because?

pmeenan:
It is not correct because easy mode needs the profiles file to work which is why it is inside the if block:

/settings/profiles.ini

PR #982, opened 2017-11-07
Update details.php

jirawat050:


www/details.php
```diff
@@ -359,9 +359,7 @@ function handleRequestHash() {
                 stepNum = 1;
                 doExpandAll = true;
             }
-            if (stepNum <= 0) {
```
pmeenan: Why did you remove this?

pmeenan: Why did you remove this?

jirawat050: sorry i Hit wrong

jirawat050:
i do change to finish

PR #983, opened 2017-11-08
Update index.php

jirawat050:
edit code before PR


www/index.php
```diff
@@ -145,6 +145,7 @@
                       echo "<li class=\"easy_mode\"><a href=\"/easy.php\">Simple Testing</a></li>";
                     }
                     ?>
+                    <li class="easy_mode"><a href="/easy.php">Simple Testing</a></li>
```
pmeenan: Easy mode only works if there is a profiles.ini in the settings directory and is already handled by th "if" case above.

Issue #990, opened 2017-11-17
Firefox potentially downloads Tracking Protection list during page load


digitarald:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1418013#c7

soulgalore:
We still use the Windows version since we haven't got Linux to work as we want on AWS, so checking the prefs (maybe) we are missing `privacy.trackingprotection.enabled false` (it's there in wpt agent and in https://wiki.mozilla.org/Security/Tracking_protection (if that's the right way to turn it off) but not in the Windows version.

pmeenan:
I disabled it in [wptagent yesterday](https://github.com/WPO-Foundation/wptagent/commit/bc045db617fc7be0c8ddb574fd314f97b8beec34) and in [wptdriver today](https://github.com/WPO-Foundation/webpagetest/commit/4c4a77f33ae303a844e20d10af16d4f9d180a97a)

wptagent already had privacy.trackingprotection.enabled set to false and was still seeing the requests so privacy.trackingprotection.annotate_channels was certainly needed (or one of the other settings - I threw the kitchen sink at it).

Please re-open if you see them come back.

PR #991, opened 2017-11-17
Trying to disable tracking protection for FF 

soulgalore:
Followed WPT agent but `privacy.trackingprotection.enabled: false` should be enough?

See #990


pmeenan:
I [just pulled all of the prefs over from wptagent](https://github.com/WPO-Foundation/webpagetest/commit/4c4a77f33ae303a844e20d10af16d4f9d180a97a) and also put a hard-block on the tracking-protection domain.

wptdriver 400 is up on the server now with the change and should be picked up by all agents over the next 1-2 hours (if your server is set to auto-update the agent).

PR #1005, opened 2017-12-03
update link and edit

Mickey248:


pmeenan:
This is an example output and is still valid, no reason to update it.

PR #1008, opened 2017-12-03
Update google_lib.inc

Mickey248:
edit link for protect


pmeenan:
Changing this to https doesn't do anything.  The protocol is just added to make it parsable.

PR #1011, opened 2017-12-03
Update Dockerfile

peterparkers:


pmeenan:
Is there a doc somewhere that says the repository should be over https?  The individual packages should be signed so it may not be adding anything and may break deb for some installs if it's not the preferred way to do it.

PR #1013, opened 2017-12-04
Update w3c_css3d.js

peterparkers:
more secure


pmeenan:
Thank you but closure compiler is an upstream project and I wouldn't want to patch it in here.  It's also in a chunk of code that is deprecated and only used at build time.

PR #1017, opened 2017-12-05
Update trace_parser.py

peterparkers:
edit http link


pmeenan:
Change should be applied to the [original project](https://github.com/WPO-Foundation/trace-parser)

PR #1022, opened 2017-12-08
Update flickr_test.html

Yulgana33:
edit link and add security


pmeenan:
That is upstream and part of the closure library.

PR #1024, opened 2017-12-08
Create html2canvas.js

jirawat050:
library to java script for dowload file pdf client


pmeenan:
If you are doing a "save as PDF" or something, can you PR the library files at the same time as the code that does the downloading?

jirawat050:
i don't pull many file  html2canvas or jspdf and i edit file comepare.php for add to work library but not pull

PR #1025, opened 2017-12-08
Create jspdf.min.js

jirawat050:
library download file pdf


pmeenan:
Is this coming with something else that uses it?

Issue #14, opened 2013-07-04
Create nice looking "executive summery" of webpagetest output


pmeenan:
Generate a document that explains what is done right and what can be improved based on a webpagetest run.  The report should be comprehendible to management, so that someone improving performance can easily show the result of their work to their boss.

Deciding what to say in this report is tricky.  Might test by showing a report to a project manager who understands the web, but is not a coder. 

https://code.google.com/p/webpagetest/issues/detail?id=17

adamwintle:
This is very interesting to me. 

We're figuring out the best way to track performance improvements (or detriments), and being able to export a report from Web Page Test would help this. A selection of export file types, such as a standard PDF or a CSV of the data which could then be imported into a Google Spreadsheet for collaboration and analysis. 

[GT Metrix](http://gtmetrix.com/)  does this very well, although unfortunately their Dashboard is single-user only and there's not an effective way to share the private tests with a team or management.

GT Metrix do however have a nice PDF report with an Executive Summary; plus an option for a Full Report with developer recommendations. Here's an example for AOI: http://gtmetrix.com/reports/www.aol.com/5xMfoIDt

giovanisp:
+1 for this.
I spent a lot of time generating reports for customers using metrics from several web performance tools. if someone is willing to code this, I am available to suggest some layouts or info. A must-have is a custom logo ;-)

adamwintle:
@giovanisp I just [opened an issue](https://github.com/WPO-Foundation/webpagetest/issues/387) to suggest a UI/UX revamp using Bootstrap; maybe we could make some print stylesheets for some report layouts?

tkadlec:
We're in the process of cleaning up issues on this project in order to ensure we're able to stay on top of high priority bugs and feature requests. 

As a part of this process, we're automatically closing any issues that have had no activity within the last two years, including this one, since the codebase has changed dramatically in that time.

If you feel this is still relevant, please [file a new issue](https://github.com/WPO-Foundation/webpagetest/issues/new/choose) using the relevant issue template so we can get it prioritized. 

Thanks!

PR #1031, opened 2017-12-09
Add function Download Filepdf for show result customwaterfall.php

pao0012:


jirawat050:
@pmeenan It correct?

pmeenan:
Do you have examples of what the output reports look like?  Grabbing an image of the current page and embedding it in a PDF doesn't seem like a strictly better result than just printing the current page to a PDF.

IF we go down this path (and it's a strong IF at this point) I'd also want the js libs to only be fetched when actually exporting.  Having to download and parse then for every visit for the few cases where someone will export a PDF is not worth the performance hit.

What I'd really expect from a proper PDF report looks more like:
- Generate an executive HTML report that includes all of the performance data, optimization suggestions (with thumbnails of the images), etc.
- Have an option to generate the same report as a PDF.
- The PDF would not be one massive image.  The text would all be text, etc.
- The PDF export could be a print-friendly stylesheet that can be used with a browser's native print to PDF support

pao0012:
Do yo need text pdf ?not image

pao0012:
i think sometime host not  english language which not support UTf8.I think to image

pao0012:
I edit not full page but to is screenshort. Sorry I edit not problem

pao0012:
i add function print .Sorry if you need

pmeenan:
Instead of submitting code for PDF exporting of current pages, please propose a design and implementation on the [existing PDF report issue](https://github.com/WPO-Foundation/webpagetest/issues/14).  Code shouldn't be written until we can agree on what PDF support should look like and how it should work.

PR #1032, opened 2017-12-09
Fixed problem in test.js

phurinaix:
previously it's not alert when you don't enter a value.

-delete && operator and three operand in first if statement.
-add if statement to check if url input equal "Host Name/IP Address" then alert.

### DIFF
--- a/www/easy.php
+++ b/www/easy.php
@@ -67,7 +67,7 @@ $profiles = parse_ini_file('./settings/profiles.ini', true);
                 </ul>
                 <div id="analytical-review" class="test_box">
                     <ul class="input_fields">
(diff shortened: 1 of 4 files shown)


pmeenan:
Not sure what you were trying to do here but a blank URL field is allowed if a script is used or a bulk url list is provided (hence the 3 operand if statement).

phurinaix:
Sorry sir
test.js has already been rectified.

-bring 3 operand back.
-add if statement to check if url input equal "Host Name/IP Address" then alert in traceroute.php.
-input of Number of Tests to Run must be a number.

I'm confused why browser didn't alert I thought the problem is bulk url.
is bulk url section hidden for client ?

phurinaix:
@pmeenan 

update test.js
-don't allow non-number input of Number of Tests to Run

PR #1033, opened 2017-12-09
Update index.php

DarkSoulCool:
Label and URL press spacebar does not.


pmeenan:
Believe this was just fixed and is not needed anymore.  Also, the label is allowed to have spaces.

PR #1034, opened 2017-12-09
Update testlog.php

DarkSoulCool:
URL press spacebar does not.


pmeenan:
Spaces are allowed in the filer (though they wouldn't really be all that useful).

PR #1035, opened 2017-12-09
edit type password and protect host check spacebar 

Mickey248:


pmeenan:
Why was the host resolver rules field removed?  Also, the password does not need to be a password field (makes it harder to make sure you passed the correct password).

Mickey248:
edit finish @pmeenan

Mickey248:
please see check correct ? @pmeenan

PR #1040, opened 2017-12-10
Update settings.ini.sample

Mickey248:
i try in update index.php because i try recaptcha to protect robot
i add site key and secret key
the picture1 i add a recaptcha api
image
the picture 2 i try API key in settings.ini
image
the picture 3 it is create recaptcha for me
image


pmeenan:
The settings.ini change should be included with the PR for the captcha code and not as a separate pull request.

There should be no php code in the settings.ini and the keys should also be shorter (and without spaces) like the rest of the settings values.

PR #1044, opened 2017-12-11
Update cdn.h

DarkSoulCool:
Edit and delete unused links


pmeenan:
umm, NO!  What are you trying to do?  That's deleting like 1/2 of the CDN detection domains.

PR #1047, opened 2017-12-12
delete all whitespace in url input

abc985:
in URL input it disable typing spacebar but it can't protect
when you copy some text that include whitespace and paste it in URL input.
I use inline javascript to protect this problem.

### DIFF
--- a/www/easy.php
+++ b/www/easy.php
@@ -67,7 +67,7 @@ $profiles = parse_ini_file('./settings/profiles.ini', true);
                 </ul>
                 <div id="analytical-review" class="test_box">
                     <ul class="input_fields">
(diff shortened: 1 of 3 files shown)


pmeenan:
It's not worth the code to do the filtering.  If there is a problem with whitespace on either end of the URL then it will [already be trimmed](https://github.com/WPO-Foundation/webpagetest/blob/master/www/runtest.php#L105) by the php.  If you are pasting urls with whitespace in the middle then you should fully expect them to fail.

If you want some level of validation I would be fine with the inputs being changed to [type="url"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url)

PR #1048, opened 2017-12-12
Update cdn.h

third1020:
Add the best cdn providers  
reference : https://www.thetoptens.com/best-content-delivery-network-cdn-providers/


pmeenan:
The limelight and MaxCDN entries don't look right.  Those are the main company domains, not the domains used for the CDN itself (and there is already detection in place for them).  Do you have a sample site that used CDNLion so I can verify the domains used there?

third1020:
Thank you for your advice

pmeenan:
Sill missing:

```Do you have a sample site that used CDNLion so I can verify the domains used there?```

PR #1052, opened 2017-12-12
Remove input quickly

fooksupachai:
I think that if user want to remove a quick input, We should use type = "search" rather than type = "text". If have any errors , I'm so sorry.

Such as =>

Before
capture1

After
capture2
 
Thank you .


pmeenan:
I appreciate the user experience of the fast clear but I'm not willing to take any side-effects that may come with it being considered a search field (which it is not).

PR #1053, opened 2017-12-12
Prevent typing space for URLs containing input

fooksupachai:
disable space key in 
- testlog.php


pmeenan:
The filter allows spaces and doesn't just filter on the URL

PR #1054, opened 2017-12-12
Update settings.ini.sample

Mickey248:
update recaptcha key for surcurity


pmeenan:
This should be a combined patch with the code and there is no way you actually tested this because the ini file format for the fields you added is just wrong.

PR #1055, opened 2017-12-12
Update traceroute.php

DarkSoulCool:
Change type text is type number


www/traceroute.php
```diff
@@ -112,7 +112,7 @@
                                 Number of Tests to Run<br>
                                 <small>Up to <?php echo $settings['maxruns']; ?></small>
                             </label>
-                            <input id="number_of_tests" type="text" class="text short" name="runs" value="3">
+                            <input id="number_of_tests" type="number" min="1" max="9" class="text short" name="runs" value="3">
```
pmeenan: The max should be set by php to the configuration setting (see index.php for an example).  It is configurable and only defaults to 9.

PR #1057, opened 2017-12-12
Fixed to Delete input video/index

DQIJAO:


pmeenan:
Why does it need an alert box in this case?  If you have a failcase where there are URLs in the list that can't be deleted it is likely a different bug.  Otherwise trying to delete an empty list should never happen.

DQIJAO:
I think it alert if nothing left input in enter data
screen shot 2017-12-13 at 9 13 40 pm

pmeenan:
That case would be better to just hide the remove link when there is only one input.  An alert box is pretty jarring and not very user-friendly.

DQIJAO:
So i should edit file video/index?

pmeenan:
Only if you know what you are doing and have a clear plan, otherwise I'm blocking all trivial PR's.

PR #1059, opened 2017-12-13
Protect spacebar

fooksupachai:
**In case the user misses, press the space bar. So, may cause system malfunction.**
  
   disable space key in 
- index.php
- traceroute.php


pmeenan:
I'm not going to be accepting any more of these js patches to filter out spaces from input whithout a REALLY good explanation of why they would be needed and why they matter for the fields being modified.

fooksupachai:
I understand you say thank you very much

PR #1060, opened 2017-12-13
add meta require for search engine

keangkai:
Tell a word about this web page and tell the details of the webpage.


pmeenan:
mozjpeg is an external library and that documentation is only used for local development.

PR #1061, opened 2017-12-13
Check Value input www/video/index.php

fooksupachai:
I think this could be an advantage. If have any errors , I'm so sorry.


pmeenan:
I am no longer accepting trivial patches like this from whatever class you are in.  If you have large, meaningful product changes or significant bug fixes I am happy to work with you on them but these small trivial patches are taking too much of my time and not meaningfully helping the project.

fooksupachai:
I understand what you are saying, Thank you very much.

PR #1062, opened 2017-12-13
add utf-8

keangkai:
add meta utf 8


pmeenan:
I am no longer accepting trivial patches like this from whatever class you are in.  If you have large, meaningful product changes or significant bug fixes I am happy to work with you on them but these small trivial patches are taking too much of my time and not meaningfully helping the project.

PR #1065, opened 2017-12-13
Update Visual comparison

qweasdzxc789:
If have 1 label and url section then remove link must be hide.

Before
image

After
picture1

### DIFF
--- a/www/video/videotest.js
+++ b/www/video/videotest.js
@@ -29,10 +29,10 @@ function RemoveUrl(id)
 }
 function showRemoveLink(){
     if (urlCount == 1){
-        $(#removeLink).hide();
+        $("#removeLink").hide();
     }
     else {
-        $(#removeLink).show();
+        $("#removeLink").show();
     }
 }
 showRemoveLink();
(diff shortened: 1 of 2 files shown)


www/video/videotest.js
```diff
@@ -21,6 +23,16 @@ function AddUrl()
 
 function RemoveUrl(id) 
 {
+    urlCount -= 1;
     $(id).remove();
     return false;
 }
+function showRemoveLink(){
```
DebugXYZ: Try to follow pattern of case for functions in the rest of the file: showRemoveLink should be ShowRemoveLink.

Also I wonder if a more appropriate name could be UpdateRemoveLink.

DebugXYZ: With styling of braces, try to follow the lead with the rest of the file (even though it may be different in other js files).

www/video/videotest.js
```diff
@@ -1,16 +1,18 @@
+var urlCount = 1;
```
DebugXYZ: I would bring the variable declaration above the first function that uses the variable, but totally a nit pick and not sure if this is just my personal preference.

tkadlec:
Lots has changed. This makes sense, but closing for a fresh start

PR #1068, opened 2017-12-14
Update mobile_devices.ini

avandonkelaar:
Added iPhone 7,8 and X


pmeenan:
7 & 8 should be the same as the 6 as far as viewport is considered.  The dimensions are for the visible region of the viewport, not the full screen size.  Is the X viewport accurate for the visible viewport in Safari?

avandonkelaar:
You're right. 6,7,and 8 are the same. The iPhone X should be different according to: 
https://www.paintcodeapp.com/news/ultimate-guide-to-iphone-resolutions. But it didn't set the dpr at 3 as it should be.

avandonkelaar:
Made the changes. Hooked up my X and verified the browsers viewport.

PR #1071, opened 2017-12-19
Docker Tweaks so container can manage AMI's eg be used in k8s cluster

timothyclarke:
These changes allow the image to be used in a [kubernetes chart](https://github.com/kubernetes/charts) so rather than just a local docker container OR an Amazon Image, one can now use the docker image to manage amazon agent instances.


pmeenan:
LGTM.  It looks like the k8s support is AWS-specific and may not work with GCE and Azure but it's great to have a starting point.  Thanks.

PR #1075, opened 2017-12-22
Update CDN List: wordpress.com/gravatar.com

josephscott:
Both wordpress.com and gravatar.com are served from a global anycast CDN.


pmeenan:
wptagent is updated as well and will take effect within the hour.  For wptdriver it will pick up the change the next time I do a build (probably not until after the holidays to keep things stable).

PR #1080, opened 2018-01-04
Add CDN detection for websites using SingularCDN

felipewd:
Added lines for either subdomain usage or server line.


felipewd:
Great, thanks!

PR #1083, opened 2018-01-09
Display versions in the 'about' page

jpvincent:
On private instances, it's never sure which version of WPT or of the agents is running.
I propose to display it in the about page.
On the .org, it would allow private instance maintainers to compare the version number to their.

### DIFF
--- a/www/about.php
+++ b/www/about.php
@@ -48,6 +48,14 @@ $page_description = "More information about WebPagetest website speed testing an
                 <?php
                 }
                 ?>
+                <p>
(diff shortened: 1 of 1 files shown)


nilskuhn:
I would appreciate to have the versions of the server and the update version for the agents on about page.
For actual agents of a running server you can find the versions already here: `<server-domain>/getTesters.php?f=html`

jpvincent:
@nilskuhn sure, this page is nice to check all of the agent version. Having it directly in /About would allow to know which version is supposed to be distributed though, if I get the mechanic correctly

@pmeenan oh I forgot the existence of this URL, thanks for reminding it :)

Well up to you to know if it's handier to have all displayed in a visible link or if WPT private instance users are supposed to be power user anyway :)

Issue #1088, opened 2018-01-14
Ignored steps for multi-step scripts


alexandrul:
Currently I'm testing a SPA application and during one of the scenarios there are a couple of steps for which the application has everything cached.

Since there are no network requests, there is no corresponding `x_y_devtools_requests.json.gz` file in the test results folder, even if there is a `x_y_devtools.json.gz` file.

Since WPT is counting the `x_y_devtools_requests.json.gz` files first, the last step(s) are not shown in the UI.

PR #1089, opened 2018-01-14
Fixed step count by looking for _devtools files first

alexandrul:
Fixes https://github.com/WPO-Foundation/webpagetest/issues/1088

### DIFF
--- a/www/include/ResultProcessing.php
+++ b/www/include/ResultProcessing.php
@@ -43,12 +43,12 @@ class ResultProcessing {
         $steps++;
       }
     }
-    // Check for agent-reportef devtools steps
+    // Check for devtools steps
     if (!$steps) {
       if ($this->cached) {
-        $pattern ="/^" . $this->run . "_Cached_([0-9]+_)?devtools_requests.json/";
+        $pattern ="/^" . $this->run . "_Cached_([0-9]+_)?devtools.json/";
       } else {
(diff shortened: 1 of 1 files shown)


alexandrul:
I'm not sure if there is a case where there could be more `_devtools_requests` files than `_devtools` ones.

If there is no such thing, the `_devtools_requests` count could be deleted.

If there is, I can update this PR to get the max from both file counts.

pmeenan:
I'll take a look at this closer in the morning but I think the best approach would be to extract the step names from any files that have them (there are 2 common patterns) and just use the highest step name found regardless of file type.  That way if anything is missing from the middle it will all still display correctly.

alexandrul:
Agree, it's a much better solution.

I would propose to discard the `cached` part in the regex, since both First View and Repeat View should have the same step count. This way, if max step number would be 5 in First View and 4 in Repeat View, the reported step count should be 5 in both cases.

pmeenan:
Pushed the fix [here](https://github.com/WPO-Foundation/webpagetest/commit/a940336c2bfe85dd1e43fa459f45567f5239f7ee).  In theory cached/FV could have different number of steps if the conditional commands are used (with the older agent that supported them).  If it becomes a display problem we should probably fix it where it is displayed.

alexandrul:
Thank you very much.

PR #1101, opened 2018-02-01
Docker php config: Do not display errors/warnings as plain pages

j0weiss:
We came across the problem when using `debug` mode on wpt server (i.e. `<wpt-server-url>/?debug=1`). Running the wpt server with docker we got the following warning messages when using starting a test run with the debug mode:
image

Since this behaviour breaks functionality we should not display the errors/warnings as plain sites. Also the user should not be confronted with the error/warning messages like this.

### DIFF
--- a/docker/server/config/php.ini
+++ b/docker/server/config/php.ini
@@ -1,3 +1,4 @@
 upload_max_filesize = 20M
 post_max_size = 20M
 memory_limit = 512M
+display_errors = false


pmeenan:
Good catch, thanks.  I'm slowly trying to clean up all of the warnings so I can remove the supression but showing them to the user is never a good idea.

PR #1123, opened 2018-03-29
Disable the chrom hook lookups which no longer work

Unterkulm5726:


pmeenan:
Was this pull request accidental?  If not, what are you trying to do?

PR #1153, opened 2018-06-26
Add hero element support

wildlyinaccurate:
Don't merge this yet! It's pretty much ready to go, pending my figuring out where to store the timing results so that they can be surfaced in resultJson.php. I'll ping you when it's fully tested and ready.

 - Allow custom element selectors to be specified on a per-test basis
 - Expose hero element timing in JSON results

Related PR: WPO-Foundation/wptagent#99

### DIFF
--- a/www/index.php
+++ b/www/index.php
@@ -161,6 +161,8 @@ $loc = ParseLocations($locations);
               echo '<input type="hidden" name="htmlbody" value="' . htmlspecialchars($_REQUEST['htmlbody']) . "\">\n";
             if (isset($_REQUEST['disable_video']))
               echo '<input type="hidden" name="disable_video" value="' . htmlspecialchars($_REQUEST['disable_video']) . "\">\n";
+            if (isset($_REQUEST['heroElementTimes']))
+              echo '<input type="hidden" name="heroElementTimes" value="' . htmlspecialchars($_REQUEST['heroElementTimes']) . "\">\n";
             ?>
 
             <h2 class="cufon-dincond_black">Test a website's performance</h2>
@@ -703,9 +705,14 @@ $loc = ParseLocations($locations);
                                             See <a href="https://sites.google.com/a/webpagetest.org/docs/using-webpagetest/custom-metrics">the documentation</a> for details on how to specify custom metrics to be captured.
                                         </div></div>
                                     </div>
-                                    
+
                                     <p><label for="custom_metrics" class="full_width">Custom Metrics:</label></p>
                                     <textarea name="custom" id="custom_metrics" cols="0" rows="0"></textarea>
+
+                                    <?php if (isset($_REQUEST['heroElementTimes']) && $_REQUEST['heroElementTimes']): ?>
+                                    <p><br><label for="hero_elements" class="full_width">Custom Hero Element Selectors:</label></p>
+                                    <textarea name="heroElements" id="hero_elements" cols="0" rows="0"></textarea>
+                                    <?php endif ?>
                                 </div>
                             </div>
 
--- a/www/index.php
+++ b/www/index.php
@@ -709,6 +709,7 @@ $loc = ParseLocations($locations);
                                     <p><label for="custom_metrics" class="full_width">Custom Metrics:</label></p>
                                     <textarea name="custom" id="custom_metrics" cols="0" rows="0"></textarea>
 
+                                    <?php if (isset($_REQUEST['heroElementTimes']) && $_REQUEST['heroElementTimes']): ?>
                                     <div class="notification-container">
                                         <br>
                                         <div class="notification"><div class="message">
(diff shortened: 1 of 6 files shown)


www/runtest.php
```diff
@@ -538,7 +538,19 @@ function DealWithMagicQuotes(&$arr) {
...
+
+            if (array_key_exists('heroElements', $_REQUEST)) {
+              $heroElements = json_decode($_REQUEST['heroElements']);
+              if (is_object($heroElements)) {
+                $heroElements = array_filter((array) $heroElements, function($selector) {
+                  // Hero elements should be an array of selector strings; filter
+                  // out invalid values.
+                  return is_string($selector) && strlen($selector);
+                });
+                $test['heroElements'] = json_encode($heroElements, JSON_FORCE_OBJECT);
```
pmeenan: I'm not 100% sure the ini file format of test jobs will like a json object.  You will probably need to base64_encode/decode like we do for ['injectScript'].

www/runtest.php
```diff
@@ -538,7 +538,19 @@ function DealWithMagicQuotes(&$arr) {
                 $test['customMetrics'][$metric] = base64_encode($code);
               }
             }
-            
+
+            if (array_key_exists('heroElements', $_REQUEST)) {
```
pmeenan: Is there a mode where the hero elements can be auto-detected (largest atf image, etc) or is the agent always going to do that independently of heroElements being specified?  I kind of expected it to find the default elements but override it if specific elements are passed in.

Also, a global setting to turn it on/off across the install and a per-test option (expecting it to add some processing overhead so not everyone may want to turn it on).  My vote would be for enabled by default with an option to turn it off.

www/page_data.inc
```diff
@@ -459,7 +459,23 @@ function loadPageStepData($localPaths, $runCompleted, $testInfo = null) {
       }
     }  
   }
-  
+
+  $hero_elements_file = $localPaths->heroElementsJsonFile();
+  if (gz_is_file($hero_elements_file)) {
+    $hero_elements = json_decode(gz_file_get_contents($hero_elements_file), true);
+    if (isset($hero_elements) && is_array($hero_elements) && isset($hero_elements['heroes'])) {
```
pmeenan: Are there also some top-level times that should be exposed.  i.e. "FirstHero" and "LastHero"?

www/runtest.php
wildlyinaccurate: There are some default elements in [hero_elements.js](https://github.com/wildlyinaccurate/wptagent/blob/visualmetrics-hero-element-times/internal/js/hero_elements.js#L24-L40) which lives on the agents. Specifically the defaults are largest H1 (with H2 as a fallback), largest IMG, largest element with a background image, and any element with the `elementtiming` attribute.

This `heroElements` option is so that you can specify selectors for hero elements at test time, in the case that you can't (or don't want to) add the `elementtiming` attribute.

wildlyinaccurate:
This is good to be reviewed whenever you're ready @pmeenan. I've done loads of testing locally, but the extent of that testing has more or less been "does this pass the options through to the agent and expose the results through jsonResult.php".

www/index.php
```diff
@@ -703,9 +705,14 @@ function onRecaptchaSubmit(token) {
                                             See <a href="https://sites.google.com/a/webpagetest.org/docs/using-webpagetest/custom-metrics">the documentation</a> for details on how to specify custom metrics to be captured.
                                         </div></div>
                                     </div>
-                                    
+
                                     <p><label for="custom_metrics" class="full_width">Custom Metrics:</label></p>
                                     <textarea name="custom" id="custom_metrics" cols="0" rows="0"></textarea>
+
+                                    <?php if (isset($_REQUEST['heroElementTimes']) && $_REQUEST['heroElementTimes']): ?>
```
pmeenan: If we are exposing it in the UI could you also create a markdown doc in the docs folder with information and examples and then link to it from here?  I'd be ok with exposing it through the API with no docs but if we add it to the UI we need to explain to people how to use it.

www/runtest.php
```diff
@@ -2170,6 +2188,10 @@ function CreateTest(&$test, $url, $batch = 0, $batch_locations = 0)
                 AddIniLine($testFile, 'lighthouse', '1');
             if( isset($test['lighthouseTrace']) && $test['lighthouseTrace'] )
                 AddIniLine($testFile, 'lighthouseTrace', '1');
+            if( $test['heroElementTimes'] )
+                AddIniLine($testFile, 'heroElementTimes', '1');
+            if( $test['heroElements'] && strlen($test['heroElements']) )
```
pmeenan: can you change the first part of this check to isset($test['heroElements']) ?

wildlyinaccurate: Thanks - done

www/index.php
wildlyinaccurate: Of course! Is a separate yellow box like this okay?
screenshot_2018-07-04 webpagetest - website performance and optimization test 1

pmeenan:
Yep. That would be perfect.

________________________________

wildlyinaccurate:
Also: Thoughts on adding Last Painted Hero to the result summary table?

screenshot from 2018-07-06 11-08-04

pmeenan:
If it will fit without pushing it out too far then I think it's worthwhile.  At a minimum it should be added to the metrics chart in the filmstrip view.  That can come as a separate PR though, merging this as it is now.

pmeenan:
Missed it in the review but ok with you if I change "First Painted Hero" and "Last Painted Hero" to not have spaces?  That will generally be easier for people to consume.

wildlyinaccurate:
Not at all! Thanks Pat :)

PR #1158, opened 2018-07-09
Expose hero times in UI

wildlyinaccurate:
I've split this out info separate commits so feel free to cherry-pick. I've exposed the hero times in a few places in the UI:

## Metrics chart in the filmstrip view (LPH)

screenshot_2018-07-09 webpagetest - visual comparison

## Test result summary table (LPH)

screenshot_2018-07-09 webpagetest test result - joseph s xps 13 www bbc com news - 07 09 18 14 14 35

## Screen shot page (all hero elements)

screenshot_2018-07-09 webpagetest screen shots - joseph s xps 13 www bbc com news - 07 09 18 14 14 35

### DIFF
--- a/www/page_data.inc
+++ b/www/page_data.inc
@@ -476,6 +476,12 @@ function loadPageStepData($localPaths, $runCompleted, $testInfo = null) {
           }
         }
       }
+
+      foreach (['FirstPaintedHero', 'LastPaintedHero'] as $metric) {
+        if (isset($ret['heroElementTimes'][$metric])) {
+          $ret[$metric] = $ret['heroElementTimes'][$metric];
+        }
+      }
     }
   }
 
--- a/www/page_data.inc
+++ b/www/page_data.inc
@@ -473,6 +473,7 @@ function loadPageStepData($localPaths, $runCompleted, $testInfo = null) {
         foreach ($hero_elements['timings'] as $timing) {
           if (isset($timing['value'])) {
             $ret['heroElementTimes'][$timing['name']] = $timing['value'];
+            $ret["heroElementTimes.{$timing['name']}"] = $timing['value'];
           }
         }
       }
--- a/www/screen_shot.php
+++ b/www/screen_shot.php
@@ -1,4 +1,4 @@
-<?php 
+<?php
 include __DIR__ . '/common.inc';
 require_once __DIR__ . '/video.inc';
 require_once __DIR__ . '/page_data.inc';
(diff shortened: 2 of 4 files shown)


www/page_data.inc
```diff
@@ -476,6 +476,12 @@ function loadPageStepData($localPaths, $runCompleted, $testInfo = null) {
           }
         }
       }
+
+      foreach (['firstPaintedHero', 'LastPaintedHero'] as $metric) {
```
pmeenan: FirstPaintedHero Needs to be capitalized.

www/screen_shot.php
```diff
@@ -230,6 +231,58 @@ function printStep($fileHandler, $testInfo, $testStepResult, $useQuicklinks) {
...
+
+                    $frame_w = imagesx($frame);
+                    $frame_h = imagesy($frame);
+                    $scale = min($frame_w / $viewport['width'], $frame_h / $viewport['height']);
+                    $x = $hero['x'] * $scale;
+                    $y = $hero['y'] * $scale;
+                    $w = $hero['width'] * $scale;
+                    $h = $hero['height'] * $scale;
+
+                    // Take a copy of the frame for the "highlight" image
```
pmeenan: Looks like this re-generates the file every time the screenshot page is accessed.  At a minimum it should check to see if the file already exists first.

That said, is there any way to do the highlighting entirely on the DOM by overlaying DIV's over the image with different transparency (and a border for the green highlight)?  That would reduce the CPU hit and storage space on the server generating images for each hero.  Probably not a big deal since the screenshot page isn't accessed that much but would like to see if it's possible first.

www/video/compare.php
```diff
@@ -722,6 +722,11 @@ function DisplayGraphs() {
     $progress_end = 0;
     foreach($tests as &$test) {
         $hasStepResult = array_key_exists('stepResult', $test) && is_a($test['stepResult'], "TestStepResult");
+        if ($hasStepResult &&
```
pmeenan: Is it worth adding all of the hero times?  Not sure how often they differ but it could be good to expose here since there is more room for metrics in this UI than the results view.

www/page_data.inc
wildlyinaccurate: Ah gee, thanks! Shoddy testing on my behalf.

www/video/compare.php
wildlyinaccurate: Yes, it might make more sense to list the individual times on this chart rather than the aggregates. Will update.

wildlyinaccurate: While this feature is still behind a flag, I'm inclined to rename the three default heroes:

* h1 -> Heading
* biggest_img -> Image
* bg_img -> BackgroundImage

That makes the names a bit friendlier in the UI, because we can just prepend "Hero" to them. "Hero Heading" reads better than "Hero h1" IMO. Thoughts?

wildlyinaccurate:
Good call on doing the element masking as an SVG, @pmeenan!

Individual hero times are now shown in the metrics chart instead of the aggregates:

screenshot_2018-07-11 webpagetest - visual comparison

PR #1161, opened 2018-07-11
Basic Authentication: type=password to password

massic80:
I think the "use test account" notice isn't enough for security: anyone accessing the browser could easily read the previously entered credentials. I'm adding a `type="password"` to the `input` field.

### DIFF
--- a/www/index.php
+++ b/www/index.php
@@ -634,7 +634,7 @@ $loc = ParseLocations($locations);
                                     </li>
                                     <li>
                                         <label for="password" style="width: auto;">Password</label>
-                                        <input type="text" name="password" id="password" class="text" style="width: 200px;">
+                                        <input type="password" name="password" id="password" class="text" style="width: 200px;">
                                     </li>
                                 </ul>
                             </div>


jakub-g:
IMO having this displayed as a regular text input instead of as a password is a feature and not a bug: _it tells you to assume that this data becomes public the moment you click "start test"._ -- because this is actually the case.

FYI, I just run a test with URL "www.example.com" and provided basic auth credentials "test"/"test".
image

When I go to WPT page of the test:
https://www.webpagetest.org/result/181011_1G_17c9b252a631266d8c1e141900e3f706/1/details/#step1_request1
...and I scroll down to "request headers" section, I can see

     authorization: Basic dGVzdDp0ZXN0

which when you base64decode it becomes `test:test`.

In other words, if someone has access to your computer, they will go to WPT "test history page" and can easily find the creds there with minimum effort.

Only use test credentials with which you're fine if they leak to the public.

massic80:
:O I didn't notice this!
I agree with you when you say _this opens your eyes on the leakage of credentials_ but I'd rather extend the issue: the authorization header (or at least the password) should be obscured on the report as well.
I come here in order to see the site's performance, not the authorization header :)

pmeenan:
Tests are private by default (and forced to be private when auth is used).  Unless you share the test result URL with someone they won't have access to it.

Scrubbing the various headers is maybe possible but between tcpdumps, netlog, traces, etc - there are just WAY too many places that credentials can be leaked.  You should treat the whole test as if the credentials will be available to anyone you share the result with.

PR #1163, opened 2018-07-11
Give hero opacity masks unique id attributes

wildlyinaccurate:
Hey Pat, I heard you like merging PRs!

Didn't realise that `id` are not scoped within an SVG, so all hero images were using the first mask. Appending the loop index to each `id` to make them unique.


pmeenan:
Your perf eval is based on number of PR's, isn't it?  ;-)

Thanks for all of your work on this.

PR #1176, opened 2018-07-25
Make webpagetest (slightly) more responsive

stuartlangridge:
A small amount of CSS overrides, applied when the page width is less than 980px, to make pages be a bit more vertical and responsive. Not intended to be a complete fix by any means, but it's a start. Done as a separate CSS file so that the changes are isolated and can be easily removed, rather than adding a media query to the 'main' CSS; eventually as the site becomes responsive overall, this would of course be integrated into the main CSS.

I don't have a full webpagetest setup here, so I can't test actual webpagetests; in particular, this may mean that these CSS changes break result pages and the like. If that's the case, then the easiest thing might be to add `class="responsive-ok"` to `<body>` in index.php and apply this CSS just to `body.responsive-ok`, and gradually extend that selector so that it covers more pages as they're fixed.

### DIFF
--- a/www/head.inc
+++ b/www/head.inc
@@ -11,6 +11,8 @@ if( strlen($page_description) )
 <link rel="shortcut icon" href="/images/favicon.ico" type="image/vnd.microsoft.icon">
 <link rel="icon" href="/images/favicon.png" type="image/png">
 
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+
 <?php 
     if(!isset($noanalytics) && extension_loaded('newrelic'))
         echo @newrelic_get_browser_timing_header();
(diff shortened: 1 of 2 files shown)


pmeenan:
Looks like it is probably a good start for handling the test settings but unless I'm missing something it doesn't work as expected:

image

stuartlangridge:
Huh, it's not doing that for me in my local version.

"Desktop":

image

"Mobile":

image

in both Firefox and Chrome. There's a bunch of stuff I don't have set up, though -- I'm just pointing a browser at the www directory, essentially -- so maybe there's something else interfering that I've missed. Not sure what, though!

pmeenan:
Could be either the sponsor or feeds section at the bottom of the page forcing the width.  I'll see if I can figure it out locally since most private instances aren't set up to do either of them.

pmeenan:
Most of the other pages don't work (filmstrip, results, etc) but I'm good with scoping it to individual pages as we convert each one.

stuartlangridge:
Yeah; it's hard to test those pages locally without having a working setup to actually _do_ tests, sadly :)

tkadlec:
👋  @stuartlangridge 

Love this, of course, but given that we have a whole new UI, I'm gonna close this one.

PR #1183, opened 2018-09-11
Preserve original user timing mark names

wildlyinaccurate:
It would be nice if we could preserve the original user timing names as well as the simplified `userTimes.{$name}` ones.

### DIFF
--- a/www/page_data.inc
+++ b/www/page_data.inc
@@ -895,6 +895,7 @@ function loadUserTimingData(&$pageData, $userTimingFile) {
               if (!isset($pageData['userTimes']))
                 $pageData['userTimes'] = array();
               $pageData['userTimes'][$name] = $time;
+              $pageData['userTimes'][$event['name']] = $time;
             }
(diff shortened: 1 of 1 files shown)


wildlyinaccurate:
Forgot to mention that measure names are already preserved, so maybe the correct change would be

```diff
-              $pageData['userTimes'][$name] = $time;
+              $pageData['userTimes'][$event['name']] = $time;
```

pmeenan:
That may work for the JSON API but you'll probably need to add escaping to the XML results and the UI if it isn't there already.

wildlyinaccurate:
Are the names escaped differently for the HTML and XML formats? This patch changes the JSON but both the HTML and XML remained the same.

I tested this using a page that has the following marks:

```js
performance.mark('foo::bar')
performance.mark('foo/bar')
performance.mark('foo "bar"')
performance.mark('foo <bar>')
performance.mark('foo bar />')
performance.mark('>>/><!-- foo bar -->')
```

PR #1187, opened 2018-09-20
Allow the "coverage" option to be passed through

wildlyinaccurate:
Seems like wptagent supports this already but it was not hooked up in webpagetest. Is this the right way to go about it? Should I add a checkbox to the UI as well?

### DIFF
--- a/www/runtest.php
+++ b/www/runtest.php
@@ -229,6 +229,8 @@
               $test['standards'] = $req_standards;
             if (isset($req_netlog))
               $test['netlog'] = $req_netlog;
+            if (isset($_REQUEST['coverage']))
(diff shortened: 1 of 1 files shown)


pmeenan:
LGTM.  I wouldn't add it to the UI until we're comfortable that it works as expected and doesn't add a bunch of overhead to the testing.  There was originally a high and low resolution way to collect JS coverage and I can't remember if the implementation to support low resolution was finished (to eliminate perf overhead).

wildlyinaccurate:
FWIW I tested a handful of pages, 9 runs each, and enabling coverage made no noticeable difference to the test duration. I didn't measure system resource usage during testing though.

PR #1189, opened 2018-09-24
Only count navigationStart as start time candidates

wildlyinaccurate:
Hey Pat! We just noticed an issue where the `chromeUserTiming` times were incorrect. I found that `ParseUserTiming()` was setting the start timestamp based on the first user timing event. However I've seen at least two test runs where the first event is InteractiveTime.

I don't think this is the intended behaviour - shouldn't the start timestamp should be based on an event that we know represents the beginning of the page navigation? This patch would limit it to the navigationStart event, but I'm not sure if there are other events that would be start time candidates.

### DIFF
--- a/www/page_data.inc
+++ b/www/page_data.inc
@@ -967,7 +967,7 @@ function ParseUserTiming($file) {
     // Find the first navigation to determine which is the main frame
     foreach ($events as $event) {
       if (is_array($event) && isset($event['name']) && isset($event['ts'])) {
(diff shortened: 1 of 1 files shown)


pmeenan:
It's complicated mostly because of multi-step scripts where a navigation may not exist.  Maybe 2 passes over the events and use one of the navigation-type events if they exist first, otherwise fall back to the first event?

I'd want to move it into the main frame detection block and allow for any of those events to be used as the start.

PR #1190, opened 2018-09-26
Fix docker build

asvinours:
Hi,

I updated the dockerfile so it builds.
```
Successfully built 2037843a4863
```

Is there an easy way to check if everything works properly in the new container?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
 FROM php:5.6-apache
 MAINTAINER iteratec WPT Team <wpt@iteratec.de>
 
-RUN echo deb http://www.deb-multimedia.org jessie main non-free >> /etc/apt/sources.list && \
-    apt-get update && \
-    DEBIAN_FRONTEND=noninteractive apt-get install -q -y --force-yes \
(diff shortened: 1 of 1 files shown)


jbuck:
@stephendonner and I will try running this in our stage environment and see if anything blows up

pmeenan:
Changes look safe and minor enough, thanks.

stephendonner:
Thanks for the (super timely for us!) patch, @asvinours!  @pmeenan + @asvinours we ran on a slightly more minimal set of changes, here: https://gist.github.com/stephendonner/7847dd936f6b36ed4b06dd5f4d06bfcd

Would you like me to submit as a PR?

Cheers

pmeenan:
If the docker build works with the smaller diff (and non-version-specific lib-png) I'll happily take a PR and merge it.  Longer-term I'd like to move it to Ubuntu bionic and Nginx and basically do the same setup as [wptserver-install](https://github.com/WPO-Foundation/wptserver-install) but that's a fair bit more work for another day.

Issue #1201, opened 2018-10-05
Dev docker container cannot be built: pecl install of xdebug fails


marlier:
Trying to install xdebug results in an error.

```
imarlier@WMF2024 ~/dev/src/webpagetest-upstream/docker/local (d3●)$ ./run_local_container.sh
Please enter your local IP adress (for XDEBUG session): 10.2.30.84
Build local docker image imarlier/webpagetest-server.
####################################################################
Sending build context to Docker daemon  6.144kB
Step 1/3 : FROM webpagetest/server
 ---> 7ed1f9d47dab
Step 2/3 : RUN yes | pecl install xdebug     && echo "zend_extension=$(find /usr/local/lib/php/extensions/ -name xdebug.so)" > /usr/local/etc/php/conf.d/xdebug.ini     && echo "xdebug.remote_enable=on" >> /usr/local/etc/php/conf.d/xdebug.ini     && echo "xdebug.remote_autostart=off" >> /usr/local/etc/php/conf.d/xdebug.ini
 ---> Running in bf5f81c7c26e
pecl/xdebug requires PHP (version >= 7.0.0), installed version is 5.6.38
No valid packages found
install failed
The command '/bin/sh -c yes | pecl install xdebug     && echo "zend_extension=$(find /usr/local/lib/php/extensions/ -name xdebug.so)" > /usr/local/etc/php/conf.d/xdebug.ini     && echo "xdebug.remote_enable=on" >> /usr/local/etc/php/conf.d/xdebug.ini     && echo "xdebug.remote_autostart=off" >> /usr/local/etc/php/conf.d/xdebug.ini' returned a non-zero code: 1
Running local container.
####################################################################
```

PR #1202, opened 2018-10-05
Fix xdebug

marlier:
Addresses #1201 (error building the dev Docker container), by forcing the installation of a version of xdebug that supports PHP < 7.0.


marlier:
Accidentally got an unrelated commit in here.  Closing, will reopen without that.

PR #1221, opened 2018-11-20
Update indicator for private tests

paulcalvano:
The test summary (when advanced settings is closed) indicates that tests are public, even though they are private by default. 

image

This change updates the `UpdateSettingsSummary` function to always indicate that tests are private in the summary.

### DIFF
--- a/www/js/test.js
+++ b/www/js/test.js
@@ -450,10 +450,7 @@ function UpdateSettingsSummary()
     if( conn != undefined )
         summary += ", " + conn.replace(/\((.)*\)/,'') + " connection";
         
-    if( $('#keep_test_private').attr('checked') )
-        summary += ", private";
-    else
-        summary += ", results are public";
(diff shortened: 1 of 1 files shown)


www/js/test.js
```diff
@@ -450,10 +450,7 @@ function UpdateSettingsSummary()
     if( conn != undefined )
         summary += ", " + conn.replace(/\((.)*\)/,'') + " connection";
         
-    if( $('#keep_test_private').attr('checked') )
-        summary += ", private";
-    else
-        summary += ", results are public";
+    summary += ", results are private";
```
pmeenan: It differs based on the setting (for private instances). Probably best to just remove the private/public indicator from the summary.

paulcalvano: Thanks.   Just updated to remove the public/private indicator from the summary.

PR #1230, opened 2018-12-13
added wao.io to cdnHeaderList

sevenval-admins:


pmeenan:
Is wai.io an actual CDN? It looks like it is a site optimization service that rewrites the content but doesn't run distributed edges.

I'm seeing ping times to the wai.io web server (which I'd hope was on the service) on the orders of 90-150ms depending on where I test from which isn't what you'd expect to see from a CDN.

sevenval-admins:
At the moment we operate our platform with two POPs and have plans to operate more POPs in the future. We are currently focusing on german and european customers. 

To those locations ping times are typically like:
```
Geography    Region         Location       Average Latency (ms)
Europe       West Europe    Netherlands    51 ms
Europe       North Europe   Ireland        78 ms
Europe       UK West        Cardiff        61 ms
Europe       UK South       London         60 ms
```

pmeenan:
I'm not comfortable calling it a CDN (at least for WebPageTest's definition) until there is at least a reasonably broad geographic distribution of POPs that are routed to based on proximity to the user (either anycast or DNS-based geolocation).  I'll be happy to take another look once you have expanded the footprint.

PR #1234, opened 2018-12-22
Fix a few 'Url' -> 'URL' typos

stephendonner:
@pmeenan r? for just a few string fixes?


stephendonner:
Also, I didn't intend to make any non-substantial (i.e. whitespace-only) changes, here; they were "helpfully" added by my Atom editor's settings (they look OK, but not being familiar with PHP, I'm cautious).

PR #1241, opened 2019-01-15
Minor string cleanup

stephendonner:
"Should just be minor string cleanup" - famous last words.

Just kidding, it really is - still trying to figure out if there's something I need to set/fix in Atom (line endings, whitespace...).


stephendonner:
@pmeenan r?  thx!

PR #1243, opened 2019-01-19
disable autocorrect & autocapitalize in the url input field

staabm:
e.g. iOS devices auto capitalize the „h“ in „https“ and this is not accepted by wtp.

Atm one need to make sure to correct the automatically upper-case character to not run into a validation error


pmeenan:
Thanks!

PR #1245, opened 2019-01-21
Add .fontawesome.com to CDN providers

lesleslie:
This adds the fly.io server header to CDN provider headers as well

### DIFF
--- a/agent/wpthook/cdn.h
+++ b/agent/wpthook/cdn.h
@@ -191,6 +191,7 @@ CDN_PROVIDER cdnList[] = {
   {".streamprovider.net", "Rocket CDN"},
   {".singularcdn.net.br", "Singular CDN"},
   {"myracloud.com", "Myracloud"},
+  {".fontawesome.com", "Font Awesome"},
   {"END_MARKER", "END_MARKER"}
 };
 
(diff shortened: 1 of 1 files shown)


pmeenan:
As best as I can tell, fontawesome.com doesn't run their own CDN.  use.fontawesome.com which is what their docs say to use looks like it is a CNAME to fontawesome-cdn.fonticons.netdna-cdn.com which is NetDNA and should already be detected.

Is there a different CDN domain or sample you can reference that I can check?

lesleslie:
Sorry I should have been more clear in the PR.

pro.fontawesome.com uses a different CDN than use.fontawesome.com

Request
Request URL:
https://pro.fontawesome.com/releases/v5.5.0/webfonts/fa-brands-400.woff2

Response
server:
Fly.io/0.1.1


When running webpagetest the following failure will appear:

Use a CDN for all static assets: 90/100

FAILED -
https://pro.fontawesome.com/releases/v5.5.0/webfonts/fa-brands-400.woff2
FAILED -
https://pro.fontawesome.com/releases/v5.5.0/webfonts/fa-regular-400.woff2
FAILED - https://pro.fontawesome.com/releases/v5.5.0/css/all.css

Let me know if you need any more information or if the PR needs adjustment.

Thank you,

-- les

Les Leslie
Principal / Developer
Wedgwood Web Works
les@wedgwoodwebworks.com
(406) 640-4222

pmeenan:
It looks like Fontawesome doesn't actually run the CDN and it is actually on Fly. I just [added Fly](https://github.com/WPO-Foundation/wptagent/commit/0823f889d237aa0c3bada7cac0f21deb14b143d9) to the detection for the current agent and it should be live within the next hour (this is actually the detection code for an older deprecated agent).

PR #1254, opened 2019-02-06
Fix Nexus 5X seetings

jeromelachaud:


www/settings/mobile_devices.ini
```diff
@@ -22,10 +22,10 @@ ua="Mozilla/5.0 (Linux; Android 5.1.1; Nexus 5 Build/LMY48B) AppleWebKit/537.36
...
 label="Nexus 5X"
 group=Android
-width=360
-height=511
-dpr=2.5
-throttle_cpu=5
+width=411
+height=600
+dpr=2.6
+throttle_cpu=2
```
pmeenan: Do you have benchmarks where you ran this? Last time I checked JS benchmarks, the 5X was SIGNIFICANTLY slower than the pixel (and pretty much everything else).

The dimensions look right based on the device I have on hand (guess something changed since 4.4) but I'm concerned about the performance with the faster throttle.

abarre: We didn't find a ranking based on JS benchmarks (such as [Speedometer2.0](https://browserbench.org/Speedometer2.0/)).

However, in [CPU benchmark](https://browser.geekbench.com/android-benchmarks), there is this result : 
Nexus 5 : 897
Nexus 5X : 1012
Pixel : 1626

Based on the CPU benchmark, I think that Nexus5X CPU throttling should be around 2.7.
Other metrics are found here https://material.io/tools/devices/.

I have a question regarding the height. The site says that Nexus 5X and Pixel has a height of 731px. How did you find 600 ?

abarre: I also see that iPhone 5C and iPhone 6 have cpu_throttling set to 1. It doesn't seem good compared to Android (https://browser.geekbench.com/ios-benchmarks).

pmeenan: The dimensions are for the browser viewport, not the full screen (which is why they are shorter).

As far as the 5C and 6 throttling goes, they were originally set by measuring the CPU time for parse/execute of JavaScript for cnn.com on the actual devices and the iOS devices were all faster than the VM that was used as the baseline, hence the "1" scale factors. 2.7 sounds good for the 5X given the current benchmarks out there.

PR #1255, opened 2019-02-09
Allow commandline options with path and origin

zimbabao:
Options like --unsafely-treat-insecure-origin-as-secure and
--user-data-dir takes origin and paths as values respectively.
Fix regex to allow such values.

Added / and : to regex, so that https://www.example.com and /tmp/my/path passes the regex.


pmeenan:
Can you instead add the explicit options that need path support to the second part of the regex that has a whitelist of options? Adding / and : to the regular character whitelist might open it up to a path-based attack against the agents.

zimbabao:
> Can you instead add the explicit options that need path support to the second part of the regex that has a whitelist of options? Adding / and : to the regular character whitelist might open it up to a path-based attack against the agents.

Done.

PR #1261, opened 2019-03-20
Enable hero element timing by default

wildlyinaccurate:
SpeedCurve have been capturing hero element timing with wptagent for a while now. I recently performed some analysis to figure out whether the times reported by wptagent were close enough to the times reported by our existing implementation. For around 95% of our tests, the times varied by <10%. I dug into a load of the remaining 5% of tests and we determined that most of the differences were for acceptable reasons - mostly caused by fuzziness, since our implementation works with JPEG frames.

Anyways, we're pretty confident that the implementation in wptagent is good enough that everyone can use it reliably, so I'd like to turn it on by default in webpagetest.org.

There will be an accompanying PR in wptagent.

### DIFF
--- a/www/runtest.php
+++ b/www/runtest.php
@@ -210,7 +210,7 @@
               $test['lighthouse'] = $req_lighthouse;
             $test['lighthouseTrace'] = isset($_REQUEST['lighthouseTrace']) && $_REQUEST['lighthouseTrace'] ? 1 : 0;
             $test['lighthouseThrottle'] = isset($_REQUEST['lighthouseThrottle']) && $_REQUEST['lighthouseThrottle'] ? 1 : GetSetting('lighthouseThrottle', 0);
-            $test['heroElementTimes'] = isset($_REQUEST['heroElementTimes']) ? (int) $_REQUEST['heroElementTimes'] : 1;
+            $test['heroElementTimes'] = isset($_REQUEST['heroElementTimes']) && $_REQUEST['heroElementTimes'] ? 1 : GetSetting('heroElementTimes', 0);
             if (isset($req_timeline))
               $test['timeline'] = $req_timeline;
             if (isset($_REQUEST['timeline_fps']) && $_REQUEST['timeline_fps'])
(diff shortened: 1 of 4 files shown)


www/runtest.php
```diff
@@ -210,7 +210,7 @@ function DealWithMagicQuotes(&$arr) {
               $test['lighthouse'] = $req_lighthouse;
             $test['lighthouseTrace'] = isset($_REQUEST['lighthouseTrace']) && $_REQUEST['lighthouseTrace'] ? 1 : 0;
             $test['lighthouseThrottle'] = isset($_REQUEST['lighthouseThrottle']) && $_REQUEST['lighthouseThrottle'] ? 1 : GetSetting('lighthouseThrottle', 0);
-            $test['heroElementTimes'] = isset($_REQUEST['heroElementTimes']) && $_REQUEST['heroElementTimes'] ? 1 : 0;
+            $test['heroElementTimes'] = isset($_REQUEST['heroElementTimes']) ? (int) $_REQUEST['heroElementTimes'] : 1;
```
pmeenan: Can you change this instead to operate like the "lighthouseThrottle" line above and pull the default value from the settings file? Also add a sample to settings.ini.sample

www/runtest.php
```diff
@@ -2235,8 +2235,8 @@ function CreateTest(&$test, $url, $batch = 0, $batch_locations = 0)
                 AddIniLine($testFile, 'lighthouseTrace', '1');
             if( isset($test['lighthouseThrottle']) && $test['lighthouseThrottle'] )
                 AddIniLine($testFile, 'lighthouseThrottle', '1');
-            if( isset($test['heroElementTimes']) && $test['heroElementTimes'] )
-                AddIniLine($testFile, 'heroElementTimes', '1');
+            if( isset($test['heroElementTimes']) )
```
pmeenan: No need to change this if the agent continues to have it disabled by default and have the system default controlled by settings.ini

wildlyinaccurate:
Thanks for the feedback, Pat! I think these latest changes are what you're looking for.

pmeenan:
Perfect, thanks. Merging now. I'm about to head out the door to Toronto for the Fitc conference so I'll flip the flag on the public instance when I get back on Sunday.

PR #1268, opened 2019-04-12
Show frame time next to hero elements and indicate LPH

wildlyinaccurate:
It would be nice to show some more context around the hero elements on the Screenshot page. This change notes the frame time and also indicates which element was the last painted hero. Do you think this is useful information? Is there a better way we could display it?

Screenshot_2019-04-12 WebPageTest Screenshots - Joseph's XPS 13 -pages-at-bbc-news - 04 12 19 15 13 58


pmeenan:
Looks good, thanks. I'm a bit worried people won't understand LPH but hopefully the title text helps.

PR #1272, opened 2019-05-14
Sticky filmstrips for better comparison between waterfall and visual progreess

j0weiss:
While I'm inspecting correlations between the waterfall and visual loading of a website, I often find myself scrolling down because of long waterfalls. But then the filmstrip scrolls out of the viewport and I'm loosing context of the visual state of the website at that point of time.

This feature enables a sticky view when accessing the filmstrip view via the URL query parameter `sticky=true`. When scrolling down the filmstrip sticks to the top of the viewport and the waterfall scrolls behind filmstrip.

This works for both, only one result or multiple results. When viewing more than 2 results the space underneath the filmstrips becomes quite limited. Therefore I decided to make it optional via URL query parameter.

### DIFF
--- a/www/video/compare.php
+++ b/www/video/compare.php
@@ -113,11 +113,17 @@ else
                 }
                 #videoContainer
                 {
-                    position: sticky;
-                    top: 0;
-                    z-index: 5;
                     <?php
-                        echo "background: #$bgcolor;\n";
+                    if (array_key_exists('sticky', $_GET) && strlen($_GET['sticky'])) {
+                    ?>
+                        position: sticky;
+                        top: 0;
+                        z-index: 5;
+                        <?php
+                            echo "background: #$bgcolor;\n";
+                        ?>
+                    <?php
+                    }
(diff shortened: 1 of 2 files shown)


j0weiss:
Hi,
I just wanted to ask if you had time to look over this pull request.
Below you can see the feature in action to get an impression of the look and feel. 😉 

stickyFilmstrip

nilskuhn:
I would appreciate it very much if this could be merged. It massively improves usability of timeframe view for longer waterfalls. 
And against all recommendations I still see very long waterfalls frequently ;) Just watched one with > 700 requests.

pmeenan:
Sorry about that. Thanks for the patch. We may be able to take it a step further and do it automatically in some cases or expose a pin in the UI or something but we can increment towards that and it's pretty awesome.

j0weiss:
No problem and thanks for merging. I'm happy to hear that you also think, that it's a feature with potential.

PR #1273, opened 2019-05-15
Adding pingback call to cancelled tests and tests forced complete

vibaldem:
We are reliant on the ping back feature to kick off our post processing as well as to maintain consistent state.  So for a bit of robustness, we are adding a ping back call when tests are cancelled, and when they are forced complete due to inactivity, so that a ping back is sent whenever a test "finishes".
Also adding an option for a default ping back url in the settings.


bramevo:
I'm looking for a similar feature but basically I want to ping back every status. So when a test is actually started besides the finished and cancelled pingback.
Any tips on how to do this?

PR #1280, opened 2019-06-24
Correct URL for list of known CDNs

bazzadp:
Was digging into how WPT figures out if a CDN was used and found this old broken URL and so submitting this as a suggested fix.


pmeenan:
Awesome, thanks.

PR #1283, opened 2019-07-06
Mitigation for all runs in sharded tests inadvertently being assigned…

### LINKED ISSUE #1074: Agents can get assigned all runs in sharded tests erroneously
In our private instance we've been seeing agents run all runs by themselves after the last run has been assigned.  This happened in with the legacy wptdriver and with the new wptagent.
Here's my best guess as to the cause:

In getwork.php the test info file is deleted outside of the lock, so its possible between unlocking and deleting the file, several threads that are waiting for the lock can read the file before it is deleted.  ProcessTestShard will be called, but it will fail to assign any runs since they're all already assigned.  The test info is then echoed back to the agent, but without an assigned run or index, which the agent will take as a non-sharded test.  

I have been unable to repro this on the public instance, since the maximum runs is limited to 9, unlike our private instance which also has additional large i/o which may contribute to the likeliness of this happening.

vibaldem:
… to one agent.
This occurs on our private instance for about 2% of 11k tests during very high load.
Fixes #1074

### DIFF
--- a/www/work/getwork.php
+++ b/www/work/getwork.php
@@ -342,7 +342,13 @@ function GetJob() {
         
         if ($is_json) {
           $testJson = TestToJSON($testInfo);
-          echo json_encode($testJson);
+          if(!isset($testJson['run']) && 
+	        GetSetting("shard_tests") && 
+			$testJson['type'] != 'traceroute'){
+            logTestMsg($testId,"Tried to start sharded test with no runs set");
+          } else {
+            echo json_encode($testJson);
+          }
         } else {
           echo $testInfo;
         }


pmeenan:
Have you identified how/when this can happen since it looks like right now the mitigation is to just log a message. It would be nice to solve the root cause.

pmeenan:
Also looks like it needs to add a check to make sure the agent supports sharding in addition to the server setting.

vibaldem:
I wrote what I thought was the cause in the linked issue:  https://github.com/WPO-Foundation/webpagetest/issues/1074

This was a couple of years back, but I think its still accurate.
The mitigation was to not return the job back to the agent, but yes, I agree that would be better to fix the locking behavior.  It was hard for me to repro reliably to test and fix, but I should be able to do it now.

PR #1331, opened 2020-02-07
Updating trace viewer

vibaldem:
+ Changing how data is loaded by using getgzip instead of file_get_contents


pmeenan:
Awesome, thanks.

wildlyinaccurate:
The trace viewer no longer works. I wonder if it's related to this PR?

```
ReferenceError: tr is not defined
trace.php:3701:99
ReferenceError: Polymer is not definedtrace.php:10513:5
```

pmeenan:
I'll see if I can find the details but you need to register an origin trial
that keeps polymer working in chrome 80+ until they rewrite the trace
viewer. There is a setting in settings.ini where you can put the origin
trial key.

pmeenan:
Origin trial is here:
https://developers.chrome.com/origintrials/#/view_trial/2431943798780067841

The setting is trace_viewer_token:
https://github.com/WPO-Foundation/webpagetest/blob/master/www/chrome/trace.php

wildlyinaccurate:
Thanks Pat! Got it working now :)

Issue #1339, opened 2020-03-02
[Feature Request] Inform the user of the Median Metric being used in the Performance Results


edwincoronado:
Currently when you visit a results page you are given the median runs on top of the page (see image below), however, there is nothing informing the user which metric was used to define the median. I think it'd be good UX addition to show display it on top. 

An idea could be to change `Performance Results (Median Run)` to `Performance Results (Median Run - <median metric>)` for example `Performance Results (Median Run - Speed Index)`

![image (2)](https://user-images.githubusercontent.com/7434954/75698304-1c5c6380-5c74-11ea-8770-1940123100cc.png)

PR #1346, opened 2020-04-15
Adding $median_metric to the `<h2>Performance Results...` element.

edwincoronado:
Added the median metric to the "Performance Results" element (see below). Tested it in a private Google instance. This closes https://github.com/WPO-Foundation/webpagetest/issues/1339

Screen Shot 2020-04-14 at 11 54 36 PM

### DIFF
--- a/www/result.inc
+++ b/www/result.inc
@@ -175,7 +175,7 @@ $page_description = "Website performance test result$testLabel.";
                 if ($fvMedian)
                 {
                     if ($testResults->countRuns() > 1)
-                        echo '<h2>Performance Results (Median Run)</h2>';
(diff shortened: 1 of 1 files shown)


www/result.inc
```diff
@@ -175,7 +175,7 @@ $page_description = "Website performance test result$testLabel.";
                 if ($fvMedian)
                 {
                     if ($testResults->countRuns() > 1)
-                        echo '<h2>Performance Results (Median Run)</h2>';
+                        echo '<h2>Performance Results (Median Run - ' . $median_metric . ')</h2>';
```
pmeenan, line 178: Just a small tweak to prevent XSS

```suggestion
                        echo '<h2>Performance Results (Median Run - ' . htmlspecialchars($median_metric) . ')</h2>';
```

PR #1349, opened 2020-05-04
feat: add security insights

lirantal:
## Changes

This PR extends the tests performed on websites and introduces:
* collecting HTTP security headers
* collecting and detecting vulnerable javascript libraries on the page

It also updates the UI to include a top-level security insights score, which is based on `securityheaders.com` as a generic benchmark so that scores should overall be similar with the exception of grade factors due to detecting vulnerable libraries as well.


www/security_checks.php
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/**
+ * Parse the page data to find any JavaScript vulnerabilities detected
+ * in libraries on the page
+ *
+ * @return boolean A array of vulnerabilities count detected by their severity
+ */
+function getSecurityVulnerabilitiesBySeverity($testInfo, $testRunResults) {
+  $testResults = $testRunResults->getstepResult(1);
```
lirantal, line 24: @pmeenan this part worked for me on my local env but I wasn't 100% sure that it will always work with the different settings that can be chosen by the user when doing a test. Just thought I'd ping you on it :-)

pmeenan: Yep, that should be fine. I don't think any of the UI deals with showing grades for multiple steps in a multi-step flow.

www/header.inc
```diff
@@ -397,6 +400,12 @@ function GetGrades($testInfo, $testRunResults, $firstRunResults)
     } elseif (isset($infoArray['lighthouse']) && $infoArray['lighthouse']) {
       $grades['lighthouse'] = array( 'class' => 'F', 'grade' => '?', 'description' => 'Lighthouse PWA Score');
     }
+
+    require_once('security_checks.php');
```
pmeenan, line 404: Can you make the security checks optional based on a setting that defaults to false? Something like GetSetting('checkSecurity'). That way it won't be enabled by default on private instances unless they opt-into it (which is how we handle the other external integrations).

www/header.inc
```diff
@@ -190,6 +190,9 @@ if( !strcasecmp('Test Result', $tab) && !@$nosubheader && !defined('EMBED') )
                           if ($check == 'lighthouse') {
                             $lighthouse = $urlGenerator->resultPage("lighthouse");
                             echo "<li class=\"$check\"><a href=\"$lighthouse\"><h2 class=\"{$grade['class']}\">{$grade['grade']}</h2></a>{$grade['description']}</li>";
+                          }else if ($check == 'security') {
+                            $snykUrl = "https://snyk.io/test/website-scanner/?test={$id}";
```
pmeenan, line 194: Should the snyk url be able to take a full URI for a test result so it can also be used by private instances (like SpeedCurve)?  Otherwise the checks should be behind a check for the $private instance global (in common.inc) and only enabled for the public instance.

www/settings/custom_metrics/securityHeaders.js
```diff
@@ -0,0 +1,117 @@
+const data = $WPT_REQUESTS;
+const HEADERS_SCORES = {
+  "strict-transport-security": 25,
+  "content-security-policy": 25,
+  "x-frame-options": 20,
+  "x-xss-protection": 20,
+  "x-content-type-options": 20,
+  "public-key-pins": 30,
```
pmeenan, line 8: Wasn't HPKP frowned upon for causing more harm than good and served as a major footgun for sites that tried to implement it? I guess it's possible to get 100 without it but it feels strange that it gets so many points.

www/security_checks.php
lirantal: Cool, thanks for confirming 👌

www/header.inc
lirantal: Ahh yes. I wrapped that part of the code with `GetSetting('securityInsights')` in this new commit: https://github.com/WPO-Foundation/webpagetest/pull/1349/commits/c162b28cc212bcadae63cae1469470f132a51192

I assumed that this means I don't need to move the security tests in custom_metrics/ but let me know if these still need to be moved into a security/ subfolder.

www/header.inc
lirantal: I think at the moment probably better to land this integration first as only available to the public WPT website and I'll re-visit this once we have deployed it as I also need to make internal adjustments on the snyk side to load any other URL if it's not the public instance.

So I can update my latest commit for GetSetting (https://github.com/WPO-Foundation/webpagetest/commit/c162b28cc212bcadae63cae1469470f132a51192) to also ensure we're on the public instance like this: 

```
if (GetSetting('securityInsights) && !$privateInstall) {
```

or is the GetSetting() enough to make sure that this is anyway not turned on by default for private instances?

www/settings/custom_metrics/securityHeaders.js
lirantal: Overall agree but Scott's FAQ pages which details the score here https://scotthelme.co.uk/scoring-transparency-on-securityheaders-io/ still references it. I pinged him to verify that but no response yet. There's no specific obligation to be in par with his website but I figured it's a good benchmark in-case people compare and will save them the trip to securityheaders.com for an additional check if we include the same checks and scores.

I'm happy to commit future updates if Scott updates me on any new grades :-)

lirantal:
@pmeenan I attended to all comments and there's a new small commit to wrap the calling code with GetSetting(). Let me know if any other updates are necessary.

lirantal:
@pmeenan I noticed there's an edge case in which we show no score at all because there aren't any security headers detected. I'm working on a fix for that and will ping when I'm done.

lirantal:
@pmeenan disregard the above, it's all ready now
I didn't notice that I was doing another test in the branch with the `if (GetSetting('securityInsights')) {}` update and that's why I didn't see the score rendering 😅

All ready on my part ✅

www/header.inc
pmeenan: Yes, still need to move it, otherwise those custom metrics scripts will run automatically for all private instances, adding overhead for no reason. They can just get a .security or .sample after the .js instead of moving them to a folder.

www/header.inc
pmeenan: GetSetting would be enough. If someone turns the setting on then they're just shooting themselves in the foot (or collecting the grade but without the link to the UI).

pmeenan:
lgtm. If you just rename the custom scripts (or put them in a subfolder) I can go ahead and merge and roll it out.

www/header.inc
lirantal: @pmeenan cool I'll rename them. After I add the suffix, do I need to do anything else in particular to read them though?

www/header.inc
lirantal: Ok thanks!

lirantal:
@pmeenan renamed the files to have a `.security` extension in this new commit https://github.com/WPO-Foundation/webpagetest/pull/1349/commits/2579935450545412bf520e14dcdb49243e1576c1 👍

PR #1356, opened 2020-05-31
Fix: security score ignores JS vulnerabilities in Wordpress websites

lirantal:
## Why this PR?

Websites built on Wordpress with a default (stock install) of a Wordpress app may include jQuery versions that include backported security fixes for older versions. When that happens, there's no effective way to correlate the vulnerability information because the version ranges will be different.

This causes false positives in terms of the overall security score being impacted for no reason

## The fix

If we detect that the website is built upon Wordpress we ignore the JavaScript vulnerabilities that were found on the website and do not account them for the overall score. The detailed page will include the same score as on WPT to keep consistency. The vulnerable JS libs will still be listed as there still may be a chance that the user is vulnerable, and so we'll include proper disclaimer there.

### Screenshot

For reference, the disclaimer on the updated details page when a WordPress website has been detected: 

image


pmeenan:
Thanks for the quick turn-around on this.

Issue #819, opened 2017-02-23
Best practice handling agentUpdate on AWS?


soulgalore:
Hey Pat, 
I got question about how we can handle agent updates. Right now we have our server configured with `agentUpdate=http://cdn.webpagetest.org/` and it works great.

We want have better control when updates are pushed (new browser etc) so we know when updates happens. Then we can minimize the risks that the updates triggers an alert (or rather we know it's because of an update).

If we would just disable the autoupdate and then enable it at certain times a week, how long would it be until the agent is updated? 

Or is better way to make sure updates are handled at certain times?

Best
Peter

pmeenan:
The browser, software and agent updates are all done separately.  For the agent updates, the server will check as part of the [hourly cron](https://github.com/WPO-Foundation/webpagetest/blob/master/www/cron/hourly.php#L25) which runs pretty close to the top of the hour.  If you want better control over it you would probably be better off just creating a stub php file that duplicates the AgentUpdate code and then trigger it whenever you want to apply an agent update.

The browsers (and python/support code) are configured on the agent in wptdriver.ini and the public AMI's are all configured to pull the installers from the public WPT config.  The images themselves don't have any Chrome or Firefox installed and they grab the currently published version at startup (and hourly) so the images can be static and never have to update.

If you want control over the browsers then you need to clone the ami's and modify wptdriver.ini to point to your server instead (or to not update the browsers and create new ami's each time a browser update comes out).  You'll also need to either manually update the .dat files and installers on your server or write some code to pull the files from the public wpt whenever you want to update.

soulgalore:
Thanks for the answer!

PR #1359, opened 2020-06-04
add documentation of agentUpdate config param

brntsllvn:
@pmeenan I am happy to document this process thoroughly if you can point me in the right direction.

This PR...
+ adds comments and documentation about `agentUpdate` setting

!! QUESTION: removing `agentUpdate` [disables the automatic agent update](https://github.com/WPO-Foundation/wptagent/issues/98#issuecomment-380690929), but I don't see [where this value is consumed](https://github.com/search?q=org%3AWPO-Foundation+agentUpdate&type=Code). I'm expecting something like `GetSetting('agentUpdate')` or similar.

!! QUESTION: what does setting `agentUpdate` have to do with setting `apkPackages`, if anything? I ask because [you link to it in this post](https://github.com/WPO-Foundation/webpagetest/issues/819#issuecomment-282000598), but I don't see where/how `agentUpdate` is used.

!! QUESTION: Is the documentation [at the link](https://github.com/WPO-Foundation/webpagetest-docs/blob/b9c5c33562c50c9821387841ec284cd30ad538bc/user/Private%20Instances/wptdriver.md#updating-test-agents) still relevant?

- https://www.webpagetest.org/work/update/update.zip response is 404
- http://www.webpagetest.org/work/update/wptupdate.zip is 404
- inspecting source at http://www.webpagetest.org does not reveal anything with either update.zip or wptupdate.zip
- [This seems to be the most up-to-date info](https://github.com/WPO-Foundation/webpagetest/issues/819#issuecomment-282000598)...? 
- inspecting source at [apkUpdate.php](https://github.com/WPO-Foundation/webpagetest/blob/master/www/cron/apkUpdate.php) seems to reveal that the hourly job retrieves a .dat file from here: http://webpagetest.org/work/update/apk.dat
- Then the hourly job just pulls the new package using `adb` [on this line](https://github.com/WPO-Foundation/webpagetest/blob/master/www/cron/apkUpdate.php#L113)

Thank you!

### DIFF
--- a/www/settings/settings.ini.sample
+++ b/www/settings/settings.ini.sample
@@ -143,6 +143,10 @@ maxNavigateCount=20
 ; (assumes a git clone and just runs "git pull origin master" as the web user).
 ;gitUpdate=1
 
+; Automatically update test agents hourly (pulls the latest test agents from the provided server)
+; Removing this line means the the old wptdriver agent will not auto-update
+agentUpdate=http://cdn.webpagetest.org/
+
 ; If an android device is connected to the server, scrape updated APK's off of the
 ; device to be installed on the test agents automatically (for cases where
 ; play auto-updating doesn't work - i.e. reverse-tethered devices)


pmeenan:
That's no longer applicable. The old wptdriver agent code hasn't been supported in a long time. If it's still in the ToT docs it should probably be entirely removed.

brntsllvn:
ok, got it. I'll find any instances and submit PRs to remove them.

PR #1360, opened 2020-06-08
Add options to pass lighthouseConfig through

wildlyinaccurate:
The matching PR to https://github.com/WPO-Foundation/wptagent/pull/352


wildlyinaccurate:
Hmm, I just realised the base64 encoding isn't automatically decoded by the agent. I thought there was some magic going on there, since so many of the other options are encoded. @pmeenan what's the correct way to pass this through to the agent?

wildlyinaccurate:
Ahh, found it. The magic is in getwork.php :)

PR #1362, opened 2020-06-09
Retrieve API key usage and limit for proactive monitoring

brntsllvn:
What's the point of this PR?
+ Background: We have hundreds of WebPageTest tests executing on schedules. These tests fail en-masse when an API key exceeds its limit.
+ Motivation: We would like to expose key usage vs. limit to proactively address keys approaching their limit.

About this PR:
+ Adds an endpoint "/getKeyUsage.php" only available in private instances. 
+ Response type JSON
+ "/getKeyUsage.php" expects an API key
+ Responds with the provided key's usage for the day and the limit

Verified in our private instance.

Feedback much appreciated.

### DIFF
--- a/www/getKeyUsage.php
+++ b/www/getKeyUsage.php
@@ -1,0 +1,69 @@
+<?php
+    header('Content-type:application/json;charset=utf-8');
+
+    require_once('common.inc');
+    if (!$privateInstall) {
+        echo FormatResult(false, "/getKeyUsage.php only available on private instances.");
+        return;
+    }
+
+    if (isset($_REQUEST['k']) && preg_match('/^(?P<key>[0-9A-Za-z]+)$/', $_REQUEST['k'], $matches)) {
+        $api_key = $matches[0];
+    } else {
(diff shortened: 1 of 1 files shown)


pmeenan:
The /usage.php entry point already provides API key usage if you pass it &k=<api_key> (even on the public instance). It was historically a HTML-only interface but I just added JSON support.

https://dev.webpagetest.org/usage.php?k=xxx&f=json

It will return 5 days worth of usage for the given key so if you want the current state, just grab the first entry in the array.

pmeenan:
btw, if you would like to add better error reporting or build out the interface more, PR's are happily accepted, just work off of the usage.php entrypoint.

brntsllvn:
Awesome. Thanks. Will do.

brntsllvn:
usage.php with json (thanks!) meets our needs for monitoring, but I'm wondering if you're amenable to adding a POST endpoint that can update an API key limit...? We haven't queued up this work yet, but I figured I'd ask while we have the hood open, and gladly submit a PR with the work when I have some time. Thanks.

PR #1377, opened 2020-07-30
Increase the width and sub-elements by 100px.

Nooshu:
## Proposal
Increase the width of the WPT container (and sub-elements) by 100px (up to 1080px). 

## Reasons
1. Since the addition of Web Vitals metrics, the metrics run outside the container and are obscured (see image below, and [live example](https://www.webpagetest.org/result/200707_1M_4831962e40e3653a9831fe93488d0285/1/details/))
overflow-issue
2. Peoples viewport sizes have increased over the years and 980px is now a very restrictive size (see user research)

## User research
We can take advantage of the fact that SpeedCurve demo their LUX RUM product on WebPageTest which gives us information about WPT users viewports. The following image is looking at the last 1 months worth of user data. 

viewport-sizes-lux

The most popular viewport size is 1280 x 720, and the smallest size listed is 1263 x 578. All other listed viewports are wider. 

Images of what the 1080px change looks like for these sizes can be seen below:

## Views not impacted
Compare, filmstrip, test history, all these pages have 100% elements which this change doesn't really effect. Only the footer / top nav change on these pages. 

## Issues that still need addressing 

1. Waterfall images need to increase in width by 100px, as the maps overhang from the containers.
overhang

2. Checklist images COULD increase by 100px, but not vital.
checklist

### DIFF
--- a/www/pagestyle.css
+++ b/www/pagestyle.css
@@ -16,8 +16,8 @@ textarea {overflow-x: hidden;width: 400px;height: 100px;padding: 3px; max-width:
 h2 .small {font-weight:normal; font-size: 12px;}
 h2 .medium {font-weight:normal; line-height: 18px; font-size: 18px;}
 .colored {color: #00A;}
-.page {width: 980px;margin: 0 auto;}
-.page-wide {width: 100%; min-width: 980px;}
+.page {width: 1080px;margin: 0 auto;}
+.page-wide {width: 100%; min-width: 1080px;}
 .centered {text-align: center;}
 .left {text-align: left;}
 .right {text-align: right;}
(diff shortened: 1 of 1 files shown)


Kartik1397:
Just curious to know, Is it possible to replace the waterfall image with html, css, js?

pmeenan:
@Kartik1397 Anything is "possible". It could be done dynamically with DOM elements or with a canvas and drawn client-side. The hard part is making sure all of the use cases for how people use the images are handled.  Copy/pasting waterfalls is really easy when it's an image and gets somewhat harder to do when it isn't.

Nooshu:
Closing this as it refers to the old UI.

PR #1383, opened 2020-08-31
fix #1204 make main metrics and W3C UserTiming tables scrollable

### LINKED ISSUE #1204: UserTiming table grows indefinitely if there's too many custom measurements
# Problem statement

If there's too many UserTiming metrics defined via `performance.mark`, and/or their names are too long and have no whitespaces, the table which contains them grows indefinitely in width and overflows. The overflow is invisible due to the constraints of the parent nodes.

https://www.webpagetest.org/result/181009_Q1_250d51ad1799abd4373099ffb16bd778/1/details/

image

The relevant code is here:
https://github.com/WPO-Foundation/webpagetest/blob/eebb39f33b49c5d65eb9e6ac0886d45e050b8fe1/www/include/UserTimingHtmlTable.php#L55-L57

With the current layout of the UserTiming table, the problem is not really solvable.

# Possible solution

One solution could be to render this table vertically instead of horizontally, i.e. 

```
before:
k1|k2|k3|...
v1|v2|v3|...

after:
k1|v1
k2|v2
k3|v3
..|..
```

This is obviously less concise than the current behavior, but I don't have a better idea.

The thing is that `#tableW3CTiming` by default already contains several items, so adding just 4-5 items is enough to make it overflow.

# Another possible semi-solution

_One "hack" to make this issue less grave is to change the CSS of `.page {width: 980px;}` which is probably a bit legacy those days - according to http://gs.statcounter.com/screen-resolution-stats/desktop/worldwide 1024x768 is less than 4% of users -- to something below `1200px` . (Note this would also help with the readability of other tables and graphs, particularly those that contains URLs shortened to some threshold using ellipses in the middle. **I opened #1207 for that**._

# Discussion

What are the opinions on this?

jakub-g:
Test URL:
https://www.webpagetest.org/result/200831_E6_4855e8cc9ddbc56bf8b40421f75ba1e9/1/details/#waterfall_view_step1

Tested in Firefox and Chrome.
image

I didn't test all the tables on all subpages, but I think they shouldn't be badly affected by this change.
But if you prefer to explicitly add scroll only to the two tables on the result page by their IDs, I could update the PR.

### DIFF
--- a/www/pagestyle.css
+++ b/www/pagestyle.css
@@ -237,7 +237,7 @@ table.result {width: 920px;}
 .bottom a:hover {text-decoration: underline;}
 
 /* shared table themes */
-table.pretty { margin-left:auto; margin-right:auto; background-color: #fff; border-collapse: collapse; border: 0px white solid; }
+table.pretty { margin-left:auto; margin-right:auto; background-color: #fff; border-collapse: collapse; border: 0px white solid; display: block; overflow-x:auto; }
 table.pretty th, table.pretty td {border: 1px silver solid; padding: 0.4em; text-align: center; }
 table.pretty th { background: gainsboro; }
 table.pretty th.empty { background: #fff; border:0; }


jakub-g:
Thanks for merging this!

PR #1392, opened 2020-09-20
Better HTTP Version parsing for HTTP Archive HAR

### LINKED ISSUE #1391: httpVersion not captured correctly in HTTP Archive har file
httpVersion is not captured correctly more often than not and seems to be blank for HTTP/1.1 requests and either `ori` or `us:`

I had a grep through the code and found this (and a similar one for response further down):

https://github.com/WPO-Foundation/webpagetest/blob/3afaeea71067efa0316786b79cae4de7d9724f8c/www/har/HttpArchiveGenerator.php#L227-L260

From my limited understanding it seems to to look for a header which has a colon (but not in first position) is a candidate for HTTP version and then grabs the 3 characters after HTTP.

While that may work in a HTTP/1.0 and HTTP/1.1 world (though even there seems flakey) it definitely won't work in a HTTP/2 which has pseudo headers which begin with a colon (so in position 0) such as `:authority:` (note the 3 characters after the 5 characters is `ori`) or `:status:` (note the 3 characters after the 5 characters is `us:`) and also only has a 1 character version number not 3.

It looks like the correct version number is in the `_protocol` column but don't think you can get a different value for request and response like I guess you could get in the pre-HTTP/2 world.

bazzadp:
Fixes #1391 

I've had a go at fixing this but not sure how to test this so could do with a code review and/or pointers how to test.

Four changes I've made (to both request and response) are:
1. Only try to parse the version from the header if not already set as first line more likely to contain the version for HTTP/1
2. Use `if ($pos !== false)` instead of `if ($pos >= 0)` as was picking up un-matched lines for request headers so setting it to blank for HTTP/1.1 (for blank lines) and `:status: 200`, `:method: GET` and `:authority: example.com` for HTTP > 1.
3. Include the full version string (`HTTP/1.1` instead of `1.1`) to be more consistent with the `protocol` field.
4. Use the `protocol` field if not able to parse (note this is same for request and response hence why I've left it to end, for those edge cases where they requests 1.0 and get back 1.1).

Things I haven't done:

- lowercase for consistency, so may be a mix of `HTTP/1.1` and `http/1.1`
- try to normalise the fields (e.g. `h2` -> `HTTP/2` or `h3-q050` -> 'HTTP/3')

Let me know if you think we should do that too.

### DIFF
--- a/www/har/HttpArchiveGenerator.php
+++ b/www/har/HttpArchiveGenerator.php
@@ -222,7 +222,7 @@ class HttpArchiveGenerator
         $request['bodySize'] = -1;
         $request['cookies'] = array();
         $request['headers'] = array();
-        $ver = '';
+        $ver = $requestData['protocol'];
         $headersSize = 0;
         if (isset($requestData['headers']) && isset($requestData['headers']['request'])) {
             foreach ($requestData['headers']['request'] as &$header) {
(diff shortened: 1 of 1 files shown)


pmeenan:
Awesome, thanks for tackling this.

PR #1440, opened 2021-03-03
Enable FCP in Firefox (and eventually, Safari)

tkadlec:
Firefox supports FCP since version 84 and Safari has it in their tech preview. We already capture the FCP metric for Firefox in the JSON, but don't expose it in the UI because we've been using the `chromeUserTiming.firstContentfulPaint` value.

This PR changes that to using the `PerformancePaintTiming.first-contentful-paint` value instead so that FCP gets exposed in the UI for non-Chrome browsers that support it.

### DIFF
--- a/www/include/RunResultHtmlTable.php
+++ b/www/include/RunResultHtmlTable.php
@@ -17,7 +17,7 @@ class RunResultHtmlTable {
   const COL_RESULT = "result";
   const COL_COST = "cost";
   const COL_CERTIFICATE_BYTES = "certificate_bytes";
-  const COL_FIRST_CONTENTFUL_PAINT = 'chromeUserTiming.firstContentfulPaint';
+  const COL_FIRST_CONTENTFUL_PAINT = 'PerformancePaintTiming.first-contentful-paint';
   const COL_LARGEST_CONTENTFUL_PAINT = 'chromeUserTiming.LargestContentfulPaint';
   const COL_CUMULATIVE_LAYOUT_SHIFT = 'chromeUserTiming.CumulativeLayoutShift';
   const COL_TOTAL_BLOCKING_TIME = 'TotalBlockingTime';
--- a/www/include/RunResultHtmlTable.php
+++ b/www/include/RunResultHtmlTable.php
@@ -17,7 +17,7 @@ class RunResultHtmlTable {
   const COL_RESULT = "result";
   const COL_COST = "cost";
   const COL_CERTIFICATE_BYTES = "certificate_bytes";
-  const COL_FIRST_CONTENTFUL_PAINT = 'PerformancePaintTiming.first-contentful-paint';
+  const COL_FIRST_CONTENTFUL_PAINT = 'FirstContentfulPaint';
   const COL_LARGEST_CONTENTFUL_PAINT = 'chromeUserTiming.LargestContentfulPaint';
   const COL_CUMULATIVE_LAYOUT_SHIFT = 'chromeUserTiming.CumulativeLayoutShift';
   const COL_TOTAL_BLOCKING_TIME = 'TotalBlockingTime';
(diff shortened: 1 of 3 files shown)


www/include/RunResultHtmlTable.php
```diff
@@ -17,7 +17,7 @@ class RunResultHtmlTable {
   const COL_RESULT = "result";
   const COL_COST = "cost";
   const COL_CERTIFICATE_BYTES = "certificate_bytes";
-  const COL_FIRST_CONTENTFUL_PAINT = 'chromeUserTiming.firstContentfulPaint';
+  const COL_FIRST_CONTENTFUL_PAINT = 'PerformancePaintTiming.first-contentful-paint';
```
pmeenan, line 20: If we switch the column we will break it for existing tests.  We should probably see if the logic that uses it can "prefer" a new value but fall back to the previous one if the new one isn't there.

Another option would be to have page_data.inc monkey-patch the value into the new location automatically.

I'd also prefer to make it a top-level "FirstContentfulPaint" and then pull it from the most accurate underlying source available (chromeUserTiming trace events will be more accurate than the in-page FCP since it accounts for JS redirects, etc).

www/page_data.inc
```diff
@@ -301,8 +301,8 @@ function loadPageStepData($localPaths, $testInfo = null) {
         $max_fid = null;
         if (isset($ret['render']) && $ret['render'] > 0)
           $seek_start = $ret['render'];
-        elseif (isset($ret['chromeUserTiming.firstContentfulPaint']) && $ret['chromeUserTiming.firstContentfulPaint'] > 0)
-          $seek_start = $ret['chromeUserTiming.firstContentfulPaint'];
+        elseif (isset($ret['PerformancePaintTiming.first-contentful-paint']) && $ret['PerformancePaintTiming.first-contentful-paint'] > 0)
+          $seek_start = $ret['PerformancePaintTiming.first-contentful-paint'];
```
pmeenan, line 305: We can change this to use the "FirstContentfulPaint" meta-value from above.

pmeenan (changes requested):

www/include/RunResultHtmlTable.php
tkadlec: > (chromeUserTiming trace events will be more accurate than the in-page FCP since it accounts for JS redirects, etc)

Was wondering that! Ok. Good call.

www/include/RunResultHtmlTable.php
```diff
@@ -17,7 +17,7 @@ class RunResultHtmlTable {
   const COL_RESULT = "result";
   const COL_COST = "cost";
   const COL_CERTIFICATE_BYTES = "certificate_bytes";
-  const COL_FIRST_CONTENTFUL_PAINT = 'PerformancePaintTiming.first-contentful-paint';
+  const COL_FIRST_CONTENTFUL_PAINT = 'FirstContentfulPaint';
```
pmeenan, line 20: Lowercase f?

www/page_data.inc
```diff
@@ -457,6 +457,15 @@ function loadPageStepData($localPaths, $testInfo = null) {
       }
     }
 
+    // set a top level firstContentfulPaint metric
```
pmeenan, line 460: This needs to be done before line 304 so the value is available for the seek start and also around 504 in case the page data is being loaded from a cached copy (for previous tests).

pmeenan (changes requested):

www/include/RunResultHtmlTable.php
tkadlec: You're too fast Meenan! :) Yeah..I fixed this.

www/page_data.inc
tkadlec: I _think_ we're in better shape here. Tested with cache on and off and getting FCP for both.

pmeenan:
Perfect, thanks

PR #1443, opened 2021-03-08
Output full object in XML response for /testStatus.php endpoint

### LINKED ISSUE #1442: Parity between JSON and XML for /testStatus.php endpoint
Currently, the testStatus.php endpoint returns more detailed test information if the JSON format is used than if the XML format is requested.

Compare: https://www.webpagetest.org/testStatus.php?test=210305_AiJ5_7788f82a41c620692bac53cef19067bc to https://www.webpagetest.org/testStatus.php?f=xml&test=210305_AiJ5_7788f82a41c620692bac53cef19067bc

Probably a fairly simple fix (famous last words).

tkadlec:
Closes #1442 by dumping the full JSON object into the XML response for the /testStatus.php endpoint. I added a little utility function to make it a bit more future-proof. All the top-level data that was there before is still there, in the same XML objects, so we should be able to more or less drop this into place.


tkadlec:
Well huh. Not sure where those other commits came from.

Look how good at Git I am, everyone.

PR #1478, opened 2021-04-14
Implementing new CLS metric

tkadlec:
With Chrome changing the way CLS is reported, figured we should update at some point, but keep around the other data because it's still important.

This PR does a couple things:

1. It moves the old CLS score to a new metric (`chromeUserTiming.TotalLayoutShift`)
2. It adds the new version of CLS as `chromeUserTiming.CumulativeLayoutShift`
3. It adds a little bit of context to the LayoutShifts object in the JSON response. Specifically, it adds:
      `window_score`: A running total of the current shift window
      `shift_window_num`: The number of the shift window an individual shift belongs in

A couple quick examples...

https://tim.webpagetest.org/jsonResult.php?test=210412_AiDcE9_b3057a7c4263ceb49f6951778301bccc&pretty=1

```json
"chromeUserTiming.TotalLayoutShift": 1.4525220889,
"chromeUserTiming.CumulativeLayoutShift": 0.41450670859999994,
"LayoutShifts": [
                    {
                        "time": 1915,
                        "score": 0.0002765496,
                        "cumulative_score": 0.0002765496,
                        "window_score": 0.0002765496,
                        "shift_window_num": 1,
                        "rects": [
                            [
                                112,
                                335,
                                1126,
                                104
                            ]
                        ],
....
```

https://tim.webpagetest.org/jsonResult.php?test=210414_BiDcRT_170b4883bf67fbfe086a56be49d807df&pretty=1

```json
"chromeUserTiming.TotalLayoutShift": 0.39808606060000007,
"chromeUserTiming.CumulativeLayoutShift": 0.2447705867,
"LayoutShifts": [
                    {
                        "time": 4025,
                        "score": 0.0045445632,
                        "cumulative_score": 0.0045445632,
                        "window_score": 0.0045445632,
                        "shift_window_num": 1,
                        "rects": [
                            [
                                52,
                                0,
                                1255,
                                40
                            ],
...
```

### DIFF
--- a/www/common.inc
+++ b/www/common.inc
@@ -152,7 +152,7 @@ if (!$privateInstall) {
 
 // constants
 define('VER_WEBPAGETEST', '20.06');   // webpagetest version
-define('VER_CSS', 119);                // version of the sitewide css file
+define('VER_CSS', 120);                // version of the sitewide css file
 define('VER_JS', 40);                 // version of the sitewide javascript file
 define('VER_JS_TEST', 45);            // version of the javascript specific to the test pages
 define('VER_JS_RUNNING', 1);          // version of the javascript specific to the test running status page
--- a/www/common.inc
+++ b/www/common.inc
@@ -152,11 +152,7 @@ if (!$privateInstall) {
 
 // constants
 define('VER_WEBPAGETEST', '20.06');   // webpagetest version
-<<<<<<< HEAD
-define('VER_CSS', 120);                // version of the sitewide css file
-=======
-define('VER_CSS', 121);                // version of the sitewide css file
->>>>>>> master
+define('VER_CSS', 122);                // version of the sitewide css file
 define('VER_JS', 40);                 // version of the sitewide javascript file
 define('VER_JS_TEST', 45);            // version of the javascript specific to the test pages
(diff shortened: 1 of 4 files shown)


tkadlec:
@pmeenan Not sure how quickly we want to implement this? Like...should we instead be adding the new version as an experimental metric for a bit?

tkadlec:
Based on

> and field tools are renaming CLS to legacy CLS or Total CLS

from https://github.com/GoogleChrome/lighthouse/issues/12350, it seems switching the legacy CLS metric to "Total" is the right call from a naming perspective.

pmeenan:
Sweet. Exactly what I was thinking of.  Can you also search for where CumulativeLayoutShift is used in the code to decide how we should update the various parts of the UI that use it?

For the filmstrip we probably want to show the window total in the denominator for each window and maybe highlight the biggest window with a different text color.

tkadlec:
Updated our layoutShiftsBeforePaint fraction to look at total (legacy) not the new cumulative.

Looking through the rest of the UI, I think it's seamlessly picking up the CLS metric which is likely what we want. 

Think filmstrip is all we've got left here.

PR #1504, opened 2021-06-16
Improve contrast of waterfall overlays

rik:
During a working session yesterday, a client expressed that it was tricky to notice the different parts of a request (bytes received vs waiting for bytes) on vitals.php. I've added a bit of blending to prop up the request details colour, that almost gives the impression that the overlay is behind the request row. And I've also applied the same changes to the regular request overlay, for consistency.


tkadlec:
Phew. That is _subtle_ but yeah, a step in the right direction. :) 

Thanks @rik!

PR #1519, opened 2021-08-16
Fix colliding default_browser_size and emulated mobile.

soulgalore:
Setting default_browser_size in the settings made all browsers (also emulated mobile) use that setting. See https://phabricator.wikimedia.org/T288970


tkadlec:
Thanks @soulgalore!

PR #1535, opened 2021-09-16
1452: Clean up custom waterfall settings form

### LINKED ISSUE #1452: Clean up custom waterfall form
The form on the custom waterfall page could use a little touch up, both to make it cleaner visually and improve the accessibility.

Screen Shot 2021-03-19 at 4 23 51 PM

scottjehl:
This page could still use some cleaning up but, some changes relevant to this particular ticket (#1452):

This update includes:
- markup updates to enclose the settings form in a details/summary, (open to any better ideas).
- some legend/fieldset updates to the markup
- CSS to display the form as a fixed panel on wider viewport sizes so that it's easier to see changes take effect as you work
- Added a download link after the waterfall image. 

Note: for the download link, I reused the existing pill style, which needed an explicit color added, hence the version bump.

fixes #1452

### DIFF
--- a/www/common.inc
+++ b/www/common.inc
@@ -149,7 +149,7 @@ if (!$privateInstall) {
 
 // constants
 define('VER_WEBPAGETEST', '21.07');   // webpagetest version
-define('VER_CSS', 142);                // version of the sitewide css file
+define('VER_CSS', 143);                // version of the sitewide css file
 define('VER_JS', 40);                 // version of the sitewide javascript file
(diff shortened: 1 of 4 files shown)


tkadlec:
This is looking great!

Can you have a quick look at it in Safari to see why the icon is off to the left? Not a huge issue, but would be nice to line that more like it is elsewhere. 

Screen Shot 2021-09-16 at 10 11 23 AM
 
When that's set, I think we're ready to merge!

tkadlec:
Looking rock solid to me! (Darn Safari)

Shipping. 🚢

PR #1540, opened 2021-09-20
1536 - offer checkbox to preserve UA when ttfb is seemingly skewed for bots

scottjehl:
Not yet ready to merge!

First pass on messaging for a checkbox that we'll offer when a result has slow ttfb and  seems to be skewed for bots. 

Live example of this message showing in a slow result: 
image
image

### DIFF
--- a/www/common.inc
+++ b/www/common.inc
@@ -149,7 +149,7 @@ if (!$privateInstall) {
 
 // constants
 define('VER_WEBPAGETEST', '21.07');   // webpagetest version
-define('VER_CSS', 143);                // version of the sitewide css file
+define('VER_CSS', 146);                // version of the sitewide css file
 define('VER_JS', 40);                 // version of the sitewide javascript file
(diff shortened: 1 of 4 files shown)


scottjehl:
Note: there's a todo in there noting that we (seemingly at least) need a public function to grab the median ttfb or other metric. There are similar functions to grab the median run number but not the data, from this particular file's access at least. I am still looking though! :)

tkadlec:
@scottjehl What do you think about only showing this message if all the test runs have a TTFB over that threshold to avoid a situation where one load really is that much slower?

If we did that, we'd need to loop through each run to get the TTFB for each OR write a function that can does that when provided with a particular metric.

scottjehl:
Oh, yeah that's even better. Good idea, thanks. I'll aim to make a function that returns an array from the runs of a particular passed metric.

tkadlec (approved):
Code comments? What is this new sorcery? ;)

tkadlec (approved):
This looks great!

PR #1566, opened 2021-10-08
Move Disable Image Format checkboxes to Advanced

tkadlec:
Thanks to https://github.com/WPO-Foundation/wptagent/pull/441, we now can use the image format checkboxes to toggle support in Firefox, so it makes sense to move this out of the "Chromium" only tab.

Most of the work was done on the agent. This moves the checkboxes and adds a little subtext to indicate which browsers the flags work in, but that's it.


jefflembeck (approved):
This all checks out. Also checked it on a test machine.

PR #1567, opened 2021-10-08
Archive and delete old, hanging, tests to avoid filling up disc space

### LINKED ISSUE #1565: Archive failed tests to avoid filling up disc [private instance]
## Feature Request Summary

Archive and delete blatantly old tests that are considered `waiting` or `running`; e.g., is the test older than `$MAX_DAYS`? Archive and Delete. `$MAX_DAYS` is defined in `archive.php` but never used (nor is it referenced anywhere else, globally).

https://github.com/WPO-Foundation/webpagetest/blob/master/www/cli/archive.php#L287

### What's the motivation or use case for the change/feature?

I run several thousand tests a day (so 10k+ runs) and need to aggresively archive them to S3 to avoid disc space filling up. I have `archive_days` set to a low number (used to be `0` but that now breaks, due to a check in `archive.php` that equates that to `false`!) - 0.1 - however I am seeing disc usage continue to increase which will result in total failure.

image

(The reason for the tests halting isn't clear - I'll raise another issue when I figure that out!)

[AB#140409](https://dev.azure.com/catchpointsystems/3530e88b-329c-4a68-a776-27327b83ff68/_workitems/edit/140409)

rposbo:
Attempts to fix #1565 

[AB#141020](https://dev.azure.com/catchpointsystems/3530e88b-329c-4a68-a776-27327b83ff68/_workitems/edit/141020)

### DIFF
--- a/www/cli/archive.php
+++ b/www/cli/archive.php
@@ -4,6 +4,7 @@
 // found in the LICENSE.md file.
 chdir('..');
 $MIN_DAYS = 2;
+$MAX_DAYS = null;
 
 require_once('common.inc');
 require_once('archive.inc');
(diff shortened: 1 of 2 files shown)


scottjehl:
This sounds like a good idea. Thanks for the PR, @rposbo 

Might want a double-check from @jefflembeck

jefflembeck (approved):
This checks out to me.

PR #1574, opened 2021-10-12
Put Layout Shift key on its own.

rik:
This saves one line vertically and I find it simpler to understand (but maybe I'm the only one).

## Before
Before

## After
After

[AB#141019](https://dev.azure.com/catchpointsystems/3530e88b-329c-4a68-a776-27327b83ff68/_workitems/edit/141019)


scottjehl:
Appreciate this idea but I think the team was undecided on whether it'll be clearer or not. I'm closing out but feel free to comment if further discussion is desired.

PR #1589, opened 2021-10-21
Add the test label to the top of test result pages

### LINKED ISSUE #1588: Add Label to Top of Test Results Page
If a label is applied to a test, we should add it to the test results summary section to make it easier to know which variation of a test you're looking at.
Screen Shot 2021-10-21 at 7 57 43 AM

[AB#141258](https://dev.azure.com/catchpointsystems/3530e88b-329c-4a68-a776-27327b83ff68/_workitems/edit/141258)

tkadlec:
Fixes #1588

Long-term, it would be nice to give this a more visual treatment and have it consistent between the test history and test results, but that styling makes it look actionable (like a button or link) and we'll have to address that later.

So, for now, simply outputting it as text.

Screen Shot 2021-10-21 at 10 44 44 AM


scottjehl:
tested! this loooks great.

PR #1617, opened 2021-11-09
feat(templates): using template for error page

jefflembeck:
This introduces the concept of a template helper.

This can be used so we better separate our logic from our templates as
we build out pages. It involves duplicating some files because I'm changing up how our header, footer, head, and hed of the main area are created.

This should make large scale changes to pages easier.

### DIFF
--- a/www/runtest.php
+++ b/www/runtest.php
@@ -1089,7 +1089,10 @@
                   $tpl = new Template('errors');
                   echo $tpl->render('runtest', array(
                     'error' => $error,
-                    'compact_mode' => $COMPACT_MODE
+                    'compact_mode' => $COMPACT_MODE,
+                    'email' => $current_user->get_email(),
+                    'supports_auth' => $supportsAuth,
+                    'supports_saml' => $supportsSaml
                   ));
                 }
             }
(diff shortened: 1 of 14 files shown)


www/runtest.php
```diff
@@ -2864,36 +2873,6 @@ function GetClosestLocation($url, $browser) {
     return $location;
 }
 
-function ErrorPage($error) {
-    global $privateInstall;
-    global $supportsAuth;
-    global $supportsSaml;
-    global $USER_EMAIL;
```
tkadlec, line 2871: `$USER_EMAIL`, `$supportsAuth` and `$supportsSaml` all get used by the header, so we need to get this information into the template somehow. (I don't see footer referencing any of these, nor do I see `$privateInstall` used at all here)

These all feel like the kind of global information we may just want to have readily available to templates throughout in some way. Does it make sense for the template class to magically pull in some basic global data or is it better to be explicit about this?

tkadlec (changes requested):

scottjehl:
👍

tkadlec (approved):
LGTM!

tkadlec:
Niiice! I like the changes. Not _crazy_ about the duplicate files, but I know that's temporary and we'll work towards fixing that up.

PR #1621, opened 2021-11-11
fix(head): add title with default layout

jefflembeck:


www/templates/layouts/default.php
```diff
@@ -6,7 +6,10 @@
 global $supportsAuth;
 global $supportsSaml;
 global $COMPACT_MODE;
+
+require_once __DIR__ . '/../../common.inc';
 ?>
+  <title><?php $page_title ? $page_title : 'WebPageTest' ?></title>
```
tkadlec, line 12: Think we're missing an echo here @jefflembeck ? 

```
<title><?php echo $page_title ? $page_title : 'WebPageTest' ?></title>
```

jefflembeck: Whew yes good call.

PR #1624, opened 2021-11-15
Fix EC2 instance lifecycle

abarre:
This commit ensures that EC2 instances are correctly set to offline before being shutdown.
I notice multiples bugs on this logic :
- EC2 instances flagged as offline was receiving tests (fix in getwork.php)
- Offline EC2 instances was not correctly reset to online when the server requires more machines (fix in ec2.inc.php on GetTesters call
- The server was terminated EC2 instances that was not offline.
- ping.php was wrongly resetting the « last work » timestamp and was blocking EC2 shutdown.


abarre:
Hello, do you need more information to evaluate this PR ?

PR #1625, opened 2021-11-15
fix(rate-limit): check runs, not tests

jefflembeck:
With the current code, somebody could potentially run 900 test runs (well,
more since the cache is in-memory) before they hit the limit. Let's
count every run.

### DIFF
--- a/www/ratelimit/check_monthly_rate_limit.php
+++ b/www/ratelimit/check_monthly_rate_limit.php
@@ -11,12 +11,15 @@ class CheckMonthlyRateLimit {
     $this->bucket = array();
   }
 
-  function check () : bool {
+  function check (?int $run_count) : bool {
+    $times_to_add = $run_count ? $run_count : 1;
     $bucket = $this->fetch_bucket();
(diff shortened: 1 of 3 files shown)


tkadlec (approved):

www/ratelimit/check_monthly_rate_limit.php
```diff
@@ -11,12 +11,15 @@ function __construct (string $ip, int $limit = 50, int $days = 28) {
...
+  function check (?int $run_count) : bool {
+    $times_to_add = $run_count ? $run_count : 1;
     $bucket = $this->fetch_bucket();
     if (count($bucket) >= $this->limit) {
       return false;
     } else {
-      $bucket[] = time();
+      for ($i = 0; $i < $times_to_add; $i++) {
+        $bucket[] = time();
+      }
```
rik, line 22:
```suggestion
array_push($bucket, ...array_fill(0, $times_to_add, time()));
```

jefflembeck: Oh that's a lot nicer!

PR #1633, opened 2021-11-30
(fix block-domains): Add a list of blocked domains to the test info header

tkadlec:
We show blocked URL patterns in our Test info summary when it applies, but not a list of blocked domains. 

This commit adds the blocked domains to the bar, like so:
Screen Shot 2021-11-30 at 10 10 38 AM

Since we have a long list of domains we block by default (hey there people trying to use webpagetest for mining crypto), we first compare the blocked domains list against our list of default blocked domains, so that we only show the specific domains that the user specified.


scottjehl (approved):
This looks good to me, @tkadlec!

PR #1635, opened 2021-12-01
fix(ga): wait until ga has loaded before firing

jefflembeck:
GA loads async, so this was never completing and thus never firing.

Also, I'm removing the IP address. We don't need it and it's
uncomfortable.

### DIFF
--- a/www/runtest.php
+++ b/www/runtest.php
@@ -3132,6 +3132,7 @@ HTML;
     }, 500);
 </script>
 HTML;
+
       $error .= loggedInPerks();
       if ($supportsSaml && $register && $apiUrl) {
         $error .= "<p>And also, if you need to run tests programmatically you might be interested in the <a href='$apiUrl'>WebPageTest API</a></p>";
--- a/www/runtest.php
+++ b/www/runtest.php
@@ -3092,7 +3092,16 @@ function CheckRateLimit($test, &$error) {
 
   if(!$passesMonthly) {
     $error = '<p>You\'ve reached the limit for logged-out tests this month, but don\'t worry! You can keep testing once you log in, which will give you access to other nice features like:</p>';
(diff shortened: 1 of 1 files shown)


www/runtest.php
```diff
@@ -3092,7 +3092,14 @@ function CheckRateLimit($test, &$error) {
...
     $error = '<p>You\'ve reached the limit for logged-out tests this month, but don\'t worry! You can keep testing once you log in, which will give you access to other nice features like:</p>';
-    $error .= '<script>window["_gaq"] && window["_gaq"].push("_trackEvent", "Error", "RateLimit", "MonthlyLimitHit", "' . $test['ip'] . '");</script>';
+    $error .= '<script>';
+    $error .= 'var intervalId = setInterval(function () { ';
+    $error .= 'if(window["_gaq"]) { ';
+    $error .= 'clearInterval(intervalId);';
+    $error .= 'window["_gaq"].push("_trackEvent", "Error", "RateLimit", "MonthlyLimitHit");';
+    $error .= ' }';
+    $error .= ' }, 500);';
+    $error .= '</script>';
```
rik, line 3102: Maybe a [HEREDOC syntax](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc) looks nicer. By using the HTML identifier, I even get HTML syntax highlighting in VSCode.

```suggestion
    $error .= <<<HTML
<script>
    var intervalId = setInterval(function () {
        if(window["_gaq"]) {
            clearInterval(intervalId);
            window["_gaq"].push("_trackEvent", "Error", "RateLimit", "MonthlyLimitHit");
        }
    }, 500);
</script>
HTML;
```

jefflembeck: _great call_

PR #1636, opened 2021-12-02
chore(deps): moving dependencies to git

jefflembeck:
Listen. I hated doing this. I truly did in every way that I can imagine,
but our deployment strategy isn't built considering things like using
composer or another package management system.

So, here we are, checking in our deps. We'll do this until we come up
with a different deployment strategy. I've got some ideas but it'll take
a bit. Maybe in the new year.


tkadlec (approved):

tkadlec:
Makes sense.

Just....no one look at what we've just done here. ;)

PR #1643, opened 2021-12-10
Fix reflected XSS issue on Filmstrip compare page

tkadlec:
Let's apply a little bit of clean-up to the $end parameter on the filmstrip page to avoid XSS issues.


scottjehl (approved):
Good change +1

jefflembeck (approved):

PR #1647, opened 2021-12-13
fix(rate-limit): Use setting for anon max

jefflembeck:


tkadlec (approved):
Yay! Looks good.

www/runtest.php
```diff
@@ -3087,7 +3087,8 @@ function CheckRateLimit($test, &$error) {
   $runcount = max(1, $test['runs']);
   $multiplier = $test['fvonly'] ? 1 : 2;
   $total_runs = $runcount * $multiplier;
-  $cmrl = new CheckMonthlyRateLimit($test['ip']);
+  $monthly_limit = GetSetting('rate_limit_anon_monthly') ? GetSetting('rate_limit_anon_monthly') : 50;
```
rik, line 3090:
```suggestion
  $monthly_limit = GetSetting('rate_limit_anon_monthly') ?: 50;
```

> It is possible to leave out the middle part of the ternary operator. Expression expr1 ?: expr3 returns expr1 if expr1 evaluates to true, and expr3 otherwise.
> -- <cite>[Ternary operator](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.ternary)</cite>

jefflembeck: Excellent catch. And a php feature I didn’t know about!

PR #1648, opened 2021-12-14
feat(design): removing compact mode

jefflembeck:
With our responsive design overhaul, compact mode doesn't mesh well (and
isn't really necessary). It also creates more code overhead to parse/work
through.


tkadlec (changes requested):
Looks good! 

Can we also kill any `.compact` styles in pagestyles2.css while we're at it? Just to clean it up there a bit too.

PR #1651, opened 2021-12-20
feat(privacy): restrict saas tests

jefflembeck:
If a test is created and run via Catchpoint's saas offerings, it should
be restricted to only those and should require a secret token to be
passed to validate that.


jefflembeck:
We're going a different route for this

PR #1663, opened 2022-01-19
feat(context): add request context to global scope

jefflembeck:
Also start autoloading and namespacing local models, I expect there to
be far more of these in the near future

Why? because I'd like to cut back on global variables and use this as a way to attach middleware-like pieces


tkadlec (approved):
Well this is looking snazzy! Nice work @jefflembeck!

PR #1666, opened 2022-01-19
feat: Adding a CP Client

jefflembeck:
Requires https://github.com/WPO-Foundation/webpagetest/pull/1665


www/models/CPClient.php
```diff
@@ -0,0 +1,72 @@
...
+  private \GuzzleHttp\Client $authClient;
+  private ?string $client_id;
+  private ?string $client_secret;
+  private ?string $grant_type;
+
+  function __construct(string $host, array $options = []) {
+    $auth_client_options = isset($options['auth_client_options']) ? $options['auth_client_options'] : array();
+    $this->client_id = isset($auth_client_options['client_id']) ? $auth_client_options['client_id'] : null;
+    $this->client_secret = isset($auth_client_options['client_secret']) ? $auth_client_options['client_secret'] : null;
+    $this->grant_type = isset($auth_client_options['grant_type']) ? $auth_client_options['grant_type'] : null;
```
rik, line 17:
```suggestion
    $auth_client_options = $options['auth_client_options'] ?? array();
    $this->client_id = $auth_client_options['client_id'] ?? null;
    $this->client_secret = $auth_client_options['client_secret'] ?? null;
    $this->grant_type = $auth_client_options['grant_type'] ?? null;
```

jefflembeck: Good catch, thanks, fixed while working on constructor tests

PR #1691, opened 2022-01-28
Fix a typo on the "aria-labelledby" attribute

Kilian:
I noticed a small typo in an attribute of the radio buttons on the homepage, so this is a fix for that.

Accessibility overview in Polypane highlighting a typo


jefflembeck (approved):
Well that's great! Thank you!

PR #1692, opened 2022-01-28
fix: use HTTPS to link to external page

Kilian:
Link on the homepage used HTTP, and a search found another one as well.

Link on the homepage that's not using HTTPS with an overlaid arrow pointing at it.
.


jefflembeck (approved):

jefflembeck:
Sweet. Thanks!

Issue #1696, opened 2022-01-28
Better highlight metrics in the detail page that are no longer in the summary table


scottjehl:
we should get total load time and requests into the details page. Maybe in the table, and either let it scroll or perhaps adjust the type size down for details.

PR #1707, opened 2022-01-31
1696

scottjehl:
One approach to improving the details tables for #1696.


tkadlec (approved):
Good call! A few folks have pinged saying they still use these metrics, so having them on the details page makes sense.

PR #1713, opened 2022-02-01
Median run for menu nav

### LINKED ISSUE #1712: Vitals Link not using median run
## Description
The Vitals page isn't choosing the median run, but instead looks to be using the first run meaning we get a difference in the vitals we report on the summary page, versus the vitals page.

This only happens when we click the vitals link the dropdown, not when we click the link by the metrics themselves.

To reproduce, go here: https://www.webpagetest.org/result/220127_AiDcQS_8462a22a72406255db2870ccaee607d7/
The click the Vitals link the dropdown navigation and notice how the vitals metrics are different than the summary page.

tkadlec:
This fixes #1712 by ensuring we always use the median run for the menu navigation.

### DIFF
--- a/www/header.inc
+++ b/www/header.inc
@@ -404,7 +404,8 @@ if( !strcasecmp('Test Result', $tab) && (!isset($nosubheader) || !@$nosubheader)
         if( !isset($run) )
             $run = $gradeRun;
         $useFriendlyUrls = !isset($_REQUEST['end']) && FRIENDLY_URLS;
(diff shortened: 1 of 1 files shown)


scottjehl (approved):
Looks great, @tkadlec! Nice catch.

Quick check: does the thumbnail image include line up with the expected run? I think it does...

tkadlec:
Just verified and yeah, it looks like it's correct as is.

PR #1718, opened 2022-02-02
feat(privacy): removing private

jefflembeck:
NOTE: THERE ARE PLANS TO ADD BACK TESTS THAT ARE USER-SCOPED LEVEL OF
PRIVATE

Currently, WPT has two privacy settings: private and public. Public
tests once showed up on the Test History page where _anybody_ could see
them. That was fun, but it's no longer the case for us.

So, now every test we run is private.

BUT, there's a caveat to that. Our version of private just means "we
don't list the URL anymore." It does not gate test results on their user
capabilities.

This is misleading. So, since there is no difference in how we treat
public vs private tests, I'm getting rid of the concept of private tests
until we hook up tests that are actually within the authorized visibility
set of the user.

### DIFF
--- a/batchtool/wpt_batch.py
+++ b/batchtool/wpt_batch.py
@@ -74,7 +74,7 @@ def RunBatch(options):
   """Run one-off batch processing of WebpageTest testing."""
 
   test_params = {'f': 'xml',
-                 'private': 1,
+                 'private': 0,
                  'priority': 6,
                  'video': options.video,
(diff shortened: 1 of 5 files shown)


tkadlec (approved):
This makes sense. We'll have _real_ private tests soon and until then, keeping this is just confusing.

:shipit: 

(Also that branch name 💯 )

PR #1723, opened 2022-02-03
chore(format): remove carriage returns

jefflembeck:
Weird rebase issues from the codebase having a mixture of carriage returns and unix line endings are rough. Let's not do that anymore. 

Depends on #1715


tkadlec:
Whew this is a thing! :)

Looks ok. Good to merge @jefflembeck?

Issue #1736, opened 2022-02-04
No more blocked url list!


tkadlec:
Looks like we lost our blocked string/domains lists. 

Before:
![image](https://user-images.githubusercontent.com/66536/152591835-e8b773f7-bd0e-4c60-9c3b-2470a0bdb91e.png)

After:
![image](https://user-images.githubusercontent.com/66536/152591867-a7aa2898-dcd8-4ac9-af98-447f56993a6e.png)

See https://twitter.com/EthanGardner/status/1489672244339040261?s=20&t=yVq84mNR7PEjKezXyd49gA for reference

PR #1737, opened 2022-02-04
Bring back the blocked list and a few others

### LINKED ISSUE #1730: Test label is not shown on details page
## Description
Test label is no longer visible in the header of the details page which mostly makes it impossible to distinguish one test from another.

Label text is in HTML inside of `<details class="heading_details">` which is hidden using CSS:
```
.heading_details {
    font-size: 0.875em;
    display: none;
    /* TODO TODO */
}
```
Please do the TODO ;)

### Steps to reproduce
Run a test with a label (bottom of the first panel of advanced settings), observe details page, fail to find the label text on the page

### Expected result
Label is visible next to the rest of the identifying information.

### Actual result
Label is not visible

tkadlec:
Turns out a few things were all lumped together here, so I made a few changes here.

First, it brings back the list of blocked strings and blocked domains, if any were set, closing #1736. Second, it adds the connectivity information (packet loss rate, bandwidth, latency) to the more info section. Finally, it closes #1730 by checking to see if there's a test label and, if so, using it instead of the device type to identify the test.

It also complete eliminates the hidden details that was lurking about, keeping things a bit tidier.

Before:
Screen Shot 2022-02-04 at 2 01 37 PM

After:
Screen Shot 2022-02-04 at 2 03 57 PM

### DIFF
--- a/www/common_lib.inc
+++ b/www/common_lib.inc
@@ -1099,6 +1099,8 @@ function GetTestInfoHtml($includeScript = true)
     global $test;
     global $isOwner;
     global $dom;
+    global $blockDomainsString;
+    global $blockString;
     global $login;
     global $admin;
     global $privateInstall;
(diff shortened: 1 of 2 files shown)


sergeychernyshev:
Thank you, @tkadlec & @scottjehl!

bitmoji

PR #1738, opened 2022-02-04
Don't unnecessarily shard a single-run test

pmeenan:
and added a setting to disable test sharding.

The main impact this has for a single-run test is that the agent will send the first and repeat view results back together instead of incrementally.  For multi-run tests with the setting turned on it lets the agent collect the full test result for possible edge-processing before sending it back.

### DIFF
--- a/www/common_lib.inc
+++ b/www/common_lib.inc
@@ -1529,9 +1529,12 @@ function AddTestJob($location, $job, $test, $testId) {
           if ($test['runs'] > 1 && substr($job, 0, 1) == '{') {
             $json = json_decode($job, true);
           }
-          for ($run = 1; $run <= $test['runs']; $run++) {
+          $submit_count = $test['runs'];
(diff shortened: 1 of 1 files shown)


vibaldem:
Are there any considerations to make for private instances where the number of runs and the resulting size of the upload may get quite large?

pmeenan:
Unless you explicitly disable sharding, there will be no change. If you
want that many runs run on a single machine and to be remote-processed,
then you can disable sharding but you'll want to make sure the web server
and PHP can handle the biggest post that will get sent.

pmeenan:
For context, the specific use case this is setting up for is the http
archive where we run -2,000 test agents (and scaling up). The plan is to
have the agent generate the final har directly but that can only work if it
has the full test.

tkadlec (approved):
Looking solid to me!

PR #1740, opened 2022-02-04
Add Page and Request Summary CSV files

### LINKED ISSUE #1739: CSV files are missing from the Export Menu
In the Export Menu there are currently options for downloading JSON, Test Logs and HAR files. In the past there was also the ability to download CSV files.  While those links are not present in the current UI, the csv.php script still exists and works.

image

For example, if I wanted the CSV files for the test: https://webpagetest.org/result/220203_AiDcNG_ffcdea171c3bb84320fdb02c95bb226c/

Then I can download 
https://webpagetest.org/csv.php?test=220203_AiDcNG_ffcdea171c3bb84320fdb02c95bb226c for the page summary
amd https://webpagetest.org/csv.php?test=220203_AiDcNG_ffcdea171c3bb84320fdb02c95bb226c&requests=1 for the request summary.

paulcalvano:
Add Page and Request Summary CSV files to the Export menu

This fixes #1739

### DIFF
--- a/www/testinfo_command-bar.inc
+++ b/www/testinfo_command-bar.inc
@@ -70,8 +70,8 @@
                       echo "<li><a href=\"/viewlog.php?test=$id\" download>Download Test Log</a></li>";
                     if (is_file("$testPath/lighthouse.log.gz"))
                       echo "<li><a href=\"/viewlog.php?test=$id&lighthouse=1\" download>Download Lighthouse Log</a></li>";
-                    echo '<li><a href="/csv.php?test=$id">Download Page CSV</a></li>';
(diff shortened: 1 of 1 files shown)


tkadlec (changes requested):
Thanks for doing this @paulcalvano!

Both these lines will need to be wrapped in double quotes (") rather than single (') for the variable interpolation to work.

So like this:

`echo "<li><a href=\"/csv.php?test=$id\">Download Page CSV</a></li>";`

Once that's fixed, we can ship this. :)

tkadlec:
Thanks, @paulcalvano!

PR #1745, opened 2022-02-09
Give interface SVGs the business

tigt:
This is mostly SVGOMG, but also I put some extra stink on it to make a strong impression and and rewrote a lot of them to be even smaller. GitHub’s rich diff viewer should come in very handy here.

Specific notes:

- For `icon-webvitals-testing`: `letter-spacing` was previously used to make the word’s bounding box centered, but `textLength` is better-suited since it will Do The Right Thing if the font metrics of `sans-serif` are different

- For `icon-advanced-testing`: I was able to remove the `crispEdges` since the overhauled dimensions all became integers, but if that still doesn’t render correctly on some browsers it can be restored

- This also fixes a very subtle visual bug in the  data URIs for the dropdown arrow: The downward caret has tiny burrs on the right side, manifesting as a stray pixel at normal sizes.


www/images/icon-external.svg
```diff
@@ -1 +1 @@
-<svg width="15" height="15" xmlns="http://www.w3.org/2000/svg"><g fill="none" fill-rule="evenodd"><path d="M7 3.5a.5.5 0 000-1v1zM12.5 8a.5.5 0 00-1 0h1zM7 2.5H3v1h4v-1zM.5 5v7h1V5h-1zM3 14.5h7v-1H3v1zm9.5-2.5V8h-1v4h1zM10 14.5a2.5 2.5 0 002.5-2.5h-1a1.5 1.5 0 01-1.5 1.5v1zM.5 12A2.5 2.5 0 003 14.5v-1A1.5 1.5 0 011.5 12h-1zM3 2.5A2.5 2.5 0 00.5 5h1A1.5 1.5 0 013 3.5v-1zM9 .5a.5.5 0 000 1v-1zm5 .5h.5a.5.5 0 00-.5-.5V1zm-.5 5a.5.5 0 001 0h-1zM9 1.5h5v-1H9v1zm4.5-.5v5h1V1h-1z" fill="#999" fill-rule="nonzero"/><path shape-rendering="crispEdges" stroke="#999" stroke-linecap="round" d="M14 1L9 6"/></g></svg>
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" width="15" height="15" fill="none" stroke="#999" stroke-linecap="round" stroke-linejoin="round"><path d="M7 3H3a2 2 0 0 0-2 2m0 0v7a2 2 0 0 0 2 2h7a2 2 0 0 0 2-2V8M9 1h5v5m0-5L9 6" shape-rendering="crispEdges"/></svg>
```
jefflembeck, line 1: The rich diff on this one is a bit off

www/images/icon-webvitals-testing.svg
```diff
@@ -1,6 +1 @@
-<svg viewBox="0 0 42 22" xmlns="http://www.w3.org/2000/svg">
-<g stroke="#FFF" fill="none" fill-rule="evenodd">
-<rect x="1" y="1" width="40" height="20" rx="4" style="fill:rgb(211,233,255);stroke:rgb(211,233,255)"/>
-<text x="5.5" y="15" letter-spacing="1.1" style="font-family:sans-serif;font-size:12px;stroke:rgb(18,29,53);fill:rgb(18,29,53)">NEW</text>
-</g>
-</svg>
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 42 22"><rect width="41" height="21" x=".5" y=".5" fill="#d3e9ff" rx="4.5"/><text x="5.5" y="15" fill="#121d35" font-weight="700" font-family="sans-serif" font-size="12" textLength="32">NEW</text></svg>
```
jefflembeck, line 1: Rich diff on these two are pretty far off, would want design clearance

jefflembeck (changes requested):
A couple of changes to be made, but otherwise, this is fantastic. Thank you!

www/images/icon-external.svg
tigt: Ah, yep, good catch. Will fix

www/images/icon-webvitals-testing.svg
tigt: Hm, I could have sworn it looked more similar. I’ll check if this is true in-browser or something with GitHub’s rich diffs

www/images/real-experience-icon.svg
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" width="110" height="110" fill="none" stroke="url(#a)" stroke-width="3"><linearGradient id="a" x2="0" y1="29" y2="82" gradientUnits="userSpaceOnUse"><stop stop-color="#4381e5"/><stop offset="1" stop-color="#afceff"/></linearGradient><path stroke-linejoin="round" d="M13.5 77V28a2.5 2.5 0 0 1 2.5-2.5h78a2.5 2.5 0 0 1 2.5 2.5v21M77 78.5H5.5V85A2.5 2.5 0 0 0 8 87.5h69M91 89v.1"/><path stroke-width="2" d="M48 50c0 5.6 2.9 8.8 8 13 4.5-3.4 8-7 8-13a8 8 0 1 0-16 0zm5 0c0-4 6-4 6 0s-6 4-6 0z"/><rect width="27" height="47" x="77.5" y="49.5" rx="2.5"/><circle cx="91" cy="89" r=".5"/></svg>
```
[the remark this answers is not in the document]
tigt: You’re right, it is. I assumed the color was chosen to match the backdrop of the icon’s use on the homepage, because if you remove the `fill` on the `<rect>` that draws the phone shape, the lines of the computer behind it show through, so I removed that geometry behind it along with the `fill`.

However, if I squint at the usage on the homepage…

The icon over a navy blue backdrop.

The phone screen is very subtly different: `#121d35` vs. the `background-color`’s `#182437`. That could be intentional, or it could be an artifact of color-space handling between the original art program vs. browsers.

If that difference is unintentional, the current suggested replacement has the bonus of letting this icon be easily reused on any background. I guess that’s a question for design?

www/images/icon-webvitals-testing.svg
tigt: GitHub is embedding them as regular old ``, so can’t be that. Looks like I assumed exchanging the `stroke` for an increase to the `font-weight` was the same, but that was wrong. Will fix

tigt: Should look the same now

www/images/icon-external.svg
tigt: Should look the same now — the `crispEdges` was non-deterministically snapping the corner depending on various viewing factors.

jefflembeck:
@tigt we have a bit of a conflict on the logos (as they are new), then I think we'll be good to go

tigt:
Should I omit them from this PR to make it easier to merge?

jefflembeck:
> Should I omit them from this PR to make it easier to merge?

I think that’s a great idea

jefflembeck (approved):

PR #1765, opened 2022-02-23
Add Total Connection Time to the Request Details Dialog

### LINKED ISSUE #1755: Display "Total Connection Time" in Request Details Modal
## Feature Request Summary
I'd like to see the "Total Connection Time" (DNS Lookup + Initial Connection + SSL Negotiation) in the Request Details Modal to more quickly figure out what might be saved by moving a resource to an open connection.

Screen Shot 2022-02-16 at 20 19 15
Screen Shot 2022-02-16 at 20 21 58

### What's the motivation or use case for the change/feature?
I often come across (very often 3P) requests that incur extra connection costs. It would be helpful to quickly see just how long the connection took to establish.

tkadlec:
Closes #1755 by adding a Total Connection Time value to the request dialog box to make it easier see the combined impact of DNS+TCP+SSL.

Also broke up the long list of details to make it easier to scan.

Before:
Screen Shot 2022-02-23 at 1 41 34 PM

After:
Screen Shot 2022-02-23 at 1 41 56 PM

### DIFF
--- a/www/waterfall.js
+++ b/www/waterfall.js
@@ -174,30 +174,30 @@ function SelectRequest(step, request) {
             details += '<b>Request Start: </b>' + (r['load_start'] / 1000.0).toFixed(3) + ' s<br>';
         if (IsValidDuration(r['dns_ms'])) {
             details += '<b>DNS Lookup: </b>' + htmlEncode(r['dns_ms']) + ' ms<br>';
-            totalConnectionTime += parseFloat(r['dns_ms']);
+            totalConnectionTime += Number(r['dns_ms']);
         } else if( r['dns_end'] !== undefined && r['dns_start'] !== undefined && r['dns_end'] > 0 ) {
             var dnsTime = r['dns_end'] - r['dns_start'];
-            totalConnectionTime+= parseFloat(dnsTime);
+            totalConnectionTime+= Number(dnsTime);
             details += '<b>DNS Lookup: </b>' + dnsTime + ' ms<br>';
         }
         if (IsValidDuration(r['connect_ms'])) {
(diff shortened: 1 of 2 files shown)


www/waterfall.js
```diff
@@ -158,52 +160,63 @@ function SelectRequest(step, request) {
...
+            }
+        }
+        details += '<h3>Timing</h3>';
         if (r['created'] !== undefined)
             details += '<b>Discovered: </b>' + (r['created'] / 1000.0).toFixed(3) + ' s<br>';
         if (r['load_start'] !== undefined)
             details += '<b>Request Start: </b>' + (r['load_start'] / 1000.0).toFixed(3) + ' s<br>';
         if (IsValidDuration(r['dns_ms'])) {
             details += '<b>DNS Lookup: </b>' + htmlEncode(r['dns_ms']) + ' ms<br>';
+            totalConnectionTime += parseFloat(r['dns_ms']);
```
jefflembeck, line 177: Do we have a guarantee that anything we're passing into `parseFloat` is a string that can be cast to a Float? (if not, we probably want to take care of that somewhere so we don't end up with concatenating `NaN` in an error.

jefflembeck: Also, are we ever parsing fractions of milliseconds? if not, maybe cast to `Number` instead?

tkadlec: Hmm....good call. I'll switch to Number.

Also, for validation, I'm thinking a Number.isNaN(totalConnectionTime) check at the end probably suffices, eh?

jefflembeck (approved):

PR #1770, opened 2022-02-28
Collection of small improvements from the HTTP Archive work

pmeenan:
* Makes the GZip level configurable: Allows for space/cpu tradeoffs for the test data stored by the php (not the gzip fiules from the agent). Dropping the gzip level to 1 cut the time in half to write the test job file on the HA server. Defaults to 6 which is the original hard-coded value.
* Added a flag (lighthouse=0) to the HAR export API to allow for disabling the including of lighthouse data in the HAR when it isn't needed.
* Moved the daily test counter incrementing into APC using atomic incrementing instead of a locked filesystem file. This reduces the contention when lots of tests are being submitted concurrently and reduces the overhead for each test.
* Fixed an exception that would be thrown in the HAR generation code if the browser data wasn't populated.
* Added a check to not re-generate the page_data on the server if it was already processed on the agent.


jefflembeck (approved):
This checks out to me and looks _great_.

PR #1773, opened 2022-03-05
Spruce up main README

tigt:
- Sections were a mix of explanations and navigation — consolidated navigation into a top-level list
- Copyediting for succinctness and clarity
- More consistent Markdown formatting

I aimed to include the same information as the one old in a more-understandable way — let me know if I omitted something!


jefflembeck (approved):
I like this!

PR #1774, opened 2022-03-05
Update links to Navigation Timing spec

tigt:
- HTTPS
- Changed host
- Correct fragment for new version


tkadlec (approved):
Nice catch! Thanks, @tigt!

PR #1832, opened 2022-05-10
Adding a menu to toggle the median metric

### LINKED ISSUE #1831: Make Median Metric changeable
Right now, on any given test, we pull the median run using the SpeedIndex metric and use that to show the metrics. However, often folks may need to look at a different metric as their primary one. They change change the median metric with the `medianMetric` url parameter, but we can make their lives a lot easier by exposing some more metrics in a list on each result page for them to quickly toggle.

tkadlec:
Closes #1831 by adding a little menu to toggle the median metric between some popular/useful options.
Screen Shot 2022-05-10 at 3 20 36 PM


scottjehl:
awesome! lgtm

PR #1833, opened 2022-05-11
fix(acct): make billing & api key tables sortable

jefflembeck:
Bring in the recommended sortable table from the w3c so this thing is
accessible and works well cross browser and then do some slight changes
to it so it matches design better

Also separate JS for the account page in an account.js file


www/templates/layouts/account.php
```diff
@@ -16,6 +16,7 @@
   <title><?php echo $page_title; ?></title>
   <?php require_once __DIR__ . '/head.inc'; ?>
   <link href="/css/account.css" rel="stylesheet" />
+  <script src="/js/account.js"></script>
```
tkadlec, line 19: @jefflembeck Does this have to be blocking or can we defer it?

www/js/account.js
```diff
@@ -0,0 +1,423 @@
...
+
+window.SortableTable = SortableTable;
+
+}(window));
+
+/**
+ * Attach all the listeners
+ */
+(() => {
+  if (document.readyState === 'loading') {
```
tkadlec, line 384: I feel like we can safely ditch the readystate check, right? The script, whether we load it blocking or deferred should always come before the `DOMContentLoaded` event. _he says wondering if there's some edge case/bug that he's not aware of_

tkadlec (changes requested):

www/js/account.js
tkadlec: ok fair :)

PR #1834, opened 2022-05-11
Add CP RUM

tkadlec:
This one adds CP RUM to our site. Just makes sense for us to walk the walk here, but also we want this so we can see the impact of some of the stuff we do.


tkadlec:
testing now....

tkadlec:
Payload looked identical, so I'm gonna keep it wrapped we can ship the tidy way.

PR #1860, opened 2022-05-20
add report only csp policy and report endpoint

deathbearbrown:
Nightmare linter strikes again.

Loves them 4 space tabs. 

Common.inc line 62-74 is the csp header and I added a report endpoint which will create a log file.

### DIFF
--- a/www/common.inc
+++ b/www/common.inc
@@ -19,24 +19,24 @@ if (Util::getSetting('php_sessions')) {
 
 $client_error = null;
 if (Util::getSetting('cp_auth')) {
-  $client_error = $_SESSION['client_error'] ?? null;
-  unset($_SESSION['client_error']);
+    $client_error = $_SESSION['client_error'] ?? null;
+    unset($_SESSION['client_error']);
 
-  set_exception_handler(function($e) {
-    if(is_a($e, ClientException::class)) {
-     $route = $e->getRoute();
-     $message = $e->getMessage();
-     $host = Util::getSetting('host');
-     unset($_SESSION['client_error']);
-     $_SESSION['client_error'] = $message;
-     $protocol = "https";
-     $location = "{$protocol}://{$host}{$route}";
-     header('Location: ' . $location);
-     exit();
-    }
-    error_log($e->getMessage);
-    throw $e;
-  });
+    set_exception_handler(function($e) {
+        if(is_a($e, ClientException::class)) {
+            $route = $e->getRoute();
+            $message = $e->getMessage();
+            $host = Util::getSetting('host');
+            unset($_SESSION['client_error']);
+            $_SESSION['client_error'] = $message;
+            $protocol = "https";
+            $location = "{$protocol}://{$host}{$route}";
+            header('Location: ' . $location);
+            exit();
+        }
+        error_log($e->getMessage);
+        throw $e;
+    });
 }
 
 // Load local deps
--- a/www/common.inc
+++ b/www/common.inc
@@ -23,11 +23,6 @@ if (Util::getSetting('cp_auth')) {
   unset($_SESSION['client_error']);
 
   set_exception_handler(function($e) {
-    echo '<pre>';
-    var_dump($e);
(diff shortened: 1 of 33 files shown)


www/csp-violation-report-endpoint.php
```diff
@@ -0,0 +1,39 @@
...
+// We pretty print the JSON before adding it to the log file
+if ($json_data = json_decode($json_data)) {
+    $json_data = json_encode($json_data, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES);
+
+    if (!file_exists($log_file)) {
+        // Send an email
+        $message = "The following Content-Security-Policy violation occurred on " .
+            $current_domain . ":\n\n" .
+            $json_data .
+            "\n\nFurther CPS violations will be logged to the following log file, but no further email notifications will be sent until this log file is deleted:\n\n" .
```
tkadlec, line 26: nit, but "Further CSP violations" :)

www/common.inc
```diff
@@ -58,8 +58,25 @@ $request_context = new RequestContext($_REQUEST, $_SERVER);
 // Disable caching by default
 header("Cache-Control: no-store, no-cache, must-revalidate, max-age=0", true);
 
+// add CSP report only
+header("Content-Security-Policy-Report-Only: default-src 'self'; 
+script-src 'report-sample' 'self' https://g.3gl.net/jp/8627/v3.3.4/M https://js.braintreegateway.com/web/dropin/1.33.0/js/dropin.min.js https://ssl.google-analytics.com/ga.js;
```
tkadlec, line 62: I'm wondering if we should include `unsafe-inline` for both `script-src` and `style-src` for now. If we don't this is going to get _very_ noisy right away. We can focus on fixing up the other stuff first and circle back to these after the rest of the policy is mostly set.

Maybe even a separate report-only policy at that point.

tkadlec: Also, for `script-src` specifically, probably good enough to just use the origin and ditch the full path?

www/csp-violation-report-endpoint.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+// Start configure
+$log_file = dirname(__FILE__) . '/csp-violations.log';
```
jefflembeck, line 4: Let's make sure we add this to the .gitignore so we don't have any issues with our deployments

www/csp-violation-report-endpoint.php
```diff
@@ -0,0 +1,39 @@
...
+    $json_data = json_encode($json_data, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES);
+
+    if (!file_exists($log_file)) {
+        // Send an email
+        $message = "The following Content-Security-Policy violation occurred on " .
+            $current_domain . ":\n\n" .
+            $json_data .
+            "\n\nFurther CPS violations will be logged to the following log file, but no further email notifications will be sent until this log file is deleted:\n\n" .
+            $log_file;
+        mail(
```
jefflembeck, line 28: Does this use the main (blocking) process? I'm worried about having 2000 users on the site and all of these going off at the same time.

www/csp-violation-report-endpoint.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+// Start configure
+$log_file = dirname(__FILE__) . '/csp-violations.log';
+$log_file_size_limit = 1000000; // bytes - once exceeded no further entries are added
+$email_address = 'slockwood@catchpoint.com';
```
jefflembeck, line 6: oh god sue i'm so sorry

deathbearbrown: haha yeah I was going to ask.. can we use a different email

www/csp-violation-report-endpoint.php
jefflembeck: Yeah for sure

deathbearbrown:
Setting this to BLOCKED, because it requires some server configurations for logging.

PR #1877, opened 2022-05-24
Event pages proxy

tkadlec:
Proxy all event pages through to pages.webpagetest.org where they live in hubspot


jefflembeck (approved):
This checks out to me!

PR #1925, opened 2022-06-01
Change to account page to show Monthly and annual pricing

deathbearbrown:
Guess who broke their local env because they typed some garbage the linter didn't pick up?? me. it was me
Screen Shot 2022-06-01 at 12 55 01 PM
Screen Shot 2022-06-01 at 12 55 06 PM


jefflembeck (approved):
I'm into it. Thank you!

PR #1953, opened 2022-06-02
Signup mobile fixes

deathbearbrown:
There were some table layout breaks for smaller screens that are fixed in this branch.

Also removed some old css that is not used anywhere.
Screen Shot 2022-06-02 at 8 23 52 AM


jefflembeck (approved):
This is nice. Thank you.

PR #1955, opened 2022-06-02
fix up passed values to experiments from three fields

### LINKED ISSUE #1927: Cannot experiment with inline styles
## Description
When I run a custom experiment with an inline style added to the end of the head, the page doesn't render properly.

### Steps to reproduce
1. Run a test and wait for the "Opportunities & Experiments" view to complete (e.g. https://www.webpagetest.org/result/220601_BiDc1B_GE8/1/experiments/).
2. Create a custom experiment with a style element (e.g. `<style>*{color:green;}</style>`) added  to the end of the head and run it.
Screen Shot 2022-06-01 at 20 23 12

### Expected result

- Inline styles should be added to the end of the head
- The changes should appear in the markup of the test variant test result 
- The inline styling changes should be visible in the screenshots of the test variant (for the example above, text should appear green)

### Actual result

- The markup of the test variant doesn't contain the inserted style element (see https://www.webpagetest.org/video/compare.php?tests=220601_BiDcSR_GEK,220601_BiDc62_GEJ)
- The background color of the HTML element appears to render – but all other content does not.

scottjehl:
fixes #1927

### DIFF
--- a/www/runtest.php
+++ b/www/runtest.php
@@ -907,10 +907,10 @@ use WebPageTest\RateLimiter;
                             if( $recipeSansId === "swap" ){
                               $experimentSwap = $ingredients;
                               if( $ingredients[0] ){
-                                $ingredients[0] = urlencode($ingredients[0]);
+                                $ingredients[0] = rawurlencode($ingredients[0]);
                               }
                               if( $ingredients[1] ){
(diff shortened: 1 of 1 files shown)


jefflembeck (approved):
🐑  it

PR #1976, opened 2022-06-06
Add Expiration to alert

deathbearbrown:
This adds the option to include an `alert_expiration` timestamp in the settings.ini along side any alerts set there.

If there is no `alert_expiration' set then it's assumed the alert has no expiration.  If it's past the date of expiration, the alert is not shown.

### DIFF
--- a/www/header.inc
+++ b/www/header.inc
@@ -31,9 +31,9 @@ if (isset($testResults)) {
 }
 
 // For users that aren't logged in, include details about the test so it can be stored in indexdb for local history support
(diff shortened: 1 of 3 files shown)


deathbearbrown:
my linter went to down on these files so please make sure you view it with whitespace turned off!
https://github.com/WPO-Foundation/webpagetest/pull/1976/files?diff=unified&w=1

tkadlec (approved):
LGTM! :shipit:

PR #1981, opened 2022-06-06
InitialOrigin will be used to pair the setCookie with the eventual non-redirect host

scottjehl:


scottjehl:
showing in this case, an origin without www and the host and cookie picking up the eventual redirected www to use
image

www/runtest.php
```diff
@@ -839,6 +839,7 @@ function buildSelfHost($hosts){
                   } else { 
                     // the first non-redirect host is passed in from experiments
                     $hostToUse = isset( $req_initialHostNonRedirect ) ? $req_initialHostNonRedirect : '%HOST%';
+                    $originToUse = isset( $req_initialOriginNonRedirect ) ? $req_initialOriginNonRedirect : '%ORIGIN%';
```
jefflembeck, line 842:
```suggestion
                    $originToUse = $req_initialOriginNonRedirect ?? '%ORIGIN%';
```

jefflembeck: you don't have to do this, but it's a nice shorthand

jefflembeck (approved):
Logic seems sound

PR #1993, opened 2022-06-07
Cancel tests interstitial

deathbearbrown:
I dislike this fix :/ 

Instead of showing this:
Screen Shot 2022-06-07 at 2 03 12 PM

it just refreshes the page which will show the page you see when you click on your test history link to a cancelled test
Screen Shot 2022-06-07 at 2 03 41 PM


deathbearbrown:
I should note that the original functionality was always a lie. It always cancels and never can't be cancelled.

PR #2004, opened 2022-06-08
Use currentSrc for LCP on experiments if it's available

tkadlec:


www/experiments/lcp.inc
```diff
@@ -26,13 +26,8 @@
         
             if (isset($lcp['type']) && $lcp['type'] === "image") {
                 if (isset($lcp['element']['src']) || isset($lcp['element']['currentSrc'])) {
-                    $lcpSource = $lcp['element']['src'];
-                    $lcpFullSource = $lcp['element']['src'];
-                    // get actual source attribute if we can.
-                    $srcAttrVal = preg_match_all('/src\=\"([^"]+)/i', $lcp['element']['outerHTML'], $srcmatches );
```
tkadlec, line 32: @scottjehl I killed the srcAttrVal bit as I don't think we need that, but not sure if there was an edge case you ran into that I'm not aware of?

tkadlec: Ah...ok, that makes sense. I'll bring this back then.

PR #2052, opened 2022-06-14
fix(js): break caching on account js

jefflembeck:
We should automate cache busting for account js (we should do this for all of them, but this is a good starter)


deathbearbrown (approved):
YESSS THANK YOU

PR #2058, opened 2022-06-15
Fixed a bunch of issues for private instances

pmeenan:
This cleans up a bunch of errors that would fire for instances where experiments are not configured and the user object isn't set.

Also cleaned up a few other errors that vscode flagged or threw while testing.


jefflembeck (approved):

tkadlec (approved):
Thanks for doing this, @pmeenan! LGTM!

PR #2060, opened 2022-06-16
Switched Google Oauth to composer and updated composer modules

pmeenan:
* Switched to the latest Google client library using composer
* Updated the composer config to require php 8.0+ (since the code doesn't work on 7.x)
* Updated all of the external libraries that were installed through composer

The only changes outside of composer are at the end:
* Code changes to www/oauth/login.php to support the updated Google client library
* Deleted lib/oauth since the library is now autoloaded from vendor
	* The CP oauth library and integration is unchanged


www/oauth/login.php
```diff
@@ -3,19 +3,18 @@
 // Use of this source code is governed by the Polyform Shield 1.0.0 license that can be
 // found in the LICENSE.md file.
 chdir('..');
+require_once __DIR__ . '/../../vendor/autoload.php';
```
jefflembeck, line 6: This isn't necessary, since `common.inc` does this to start.

pmeenan: Whoops.  Good catch, thanks. Done.

jefflembeck (approved):
Thank you

PR #2063, opened 2022-06-17
Pro status visuals

scottjehl:
this change better communicates that you're logged in and paid for pro features via the header and homepage messaging.

screencap
image

### DIFF
--- a/www/common.inc
+++ b/www/common.inc
@@ -372,7 +372,7 @@ if (strlen($id)) {
 
       // $owner is set by CP details in AttachUser middleware if loaded
       $owner_id_matches_test = array_key_exists('owner', $test['testinfo']) && strlen($owner) && $owner == $test['testinfo']['owner'];
-      $uid_matches_test = array_key_exists('uid', $test['testinfo']) && isset($uid) && strlen($uid) && $uid == $test['testinfo']['uid'];
+      $uid_matches_test = array_key_exists('uid', $test['testinfo']) && strlen($uid) && $uid == $test['testinfo']['uid'];
 
       $isOwner = $owner_id_matches_test || $uid_matches_test;
       $test_is_private = array_key_exists('private', $test['testinfo']) && !!$test['testinfo']['private'];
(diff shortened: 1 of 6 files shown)


www/common.inc
```diff
@@ -372,7 +372,7 @@ if (strlen($id)) {
 
       // $owner is set by CP details in AttachUser middleware if loaded
       $owner_id_matches_test = array_key_exists('owner', $test['testinfo']) && strlen($owner) && $owner == $test['testinfo']['owner'];
-      $uid_matches_test = array_key_exists('uid', $test['testinfo']) && isset($uid) && strlen($uid) && $uid == $test['testinfo']['uid'];
+      $uid_matches_test = array_key_exists('uid', $test['testinfo']) && strlen($uid) && $uid == $test['testinfo']['uid'];
```
jefflembeck, line 375: Any reason for this change? I believe we need it for private installs

www/common.inc
```diff
@@ -442,7 +442,7 @@ if (strlen($id)) {
     if( $test['testinfo'] ){
       $testTestInfo = $test['testinfo'];
       $metaInfo = $testTestInfo['metadata'];
-      if( isset($metaInfo) && is_string($metaInfo) ){
+      if( $metaInfo ){
```
jefflembeck, line 445: Same as 🔴

www/home_header.php
```diff
@@ -1,33 +1,44 @@
 <div class="home_feature">
     <div class="home_feature_hed_contain">
 		<div class="home_feature_containslides">
-			<div class="home_feature_hed home_feature_hed-main">
-				<div class="home_feature_hed_text">
-					<h1 class="attention">Test. Optimize. Repeat!</h1>
-					<p>Instantly test your site’s speed, usability, and resilience in real <b>browsers</b>, <b>devices</b>, and <b>locations</b> around the&nbsp;world. </p>
+			<?php 			
 
+			if( $experiments_paid && experiments_logged_in ){
```
jefflembeck, line 6:
```suggestion
			if( $experiments_paid && $experiments_logged_in ){
```

www/runtest.php
```diff
@@ -2,7 +2,22 @@
 // Copyright 2020 Catchpoint Systems Inc.
 // Use of this source code is governed by the Polyform Shield 1.0.0 license that can be
 // found in the LICENSE.md file.
-  require_once(__DIR__ . '/../vendor/autoload.php');
+require_once(__DIR__ . '/../vendor/autoload.php');
+
+    // deal with magic quotes being enabled
```
jefflembeck, line 7: why add this back in?

www/common.inc
scottjehl: This was a mistake. Should be fixed Thank you!

www/common.inc
scottjehl: Same as above reply, thanks!

tkadlec (approved):
Think this is looking good now, Scott. 🚢

PR #2072, opened 2022-06-24
disable the rerun test button from the compare page, as we have no cu…

scottjehl:
…rrent way to rerun full experiments


scottjehl:
Tested, merging.

Issue #2076, opened 2022-06-27
Plan UI design changes for highlighting remaining run count


scottjehl:
This should be considered in Start Test (homepage), account), and eventually in test history (addressed in separate issue)

PR #2079, opened 2022-06-27
addresses messaging for #2076. 

scottjehl:
Instructions:
1. Before merging this PR, the relevant index.php markup needs to be populated with php values and made the same.
2. For sake of example, there are two variations in the markup currently which show the two possible states of the "runs left" message. These are just before the submit inputs in the SIMPLE and ADVANCED groups. Again, they should be the same once implemented in PHP.
  - When there are more than 30 runs remaining, we can show this markup just before the inputs:
    ```html
    <small class="test_runs"><span>45 Runs Left</span> | <a href="/account">Upgrade</a></small>
    ```
    image
  - When there are fewer than 30 runs left, we can change the markup slightly to add this warn class, and also add an `aria-disabled` attribute to the input that follows the `small` element:
    ```html
    <small class="test_runs test_runs-warn"><span>15 Runs Left</span> | <a href="/account">Upgrade</a></small>
    <input 
    ```
    image
    Small note on the `aria-disabled`: I've used that attr alone to allow focus to still occur, which `disabled` would otherwise block. There's CSS included to set opacity and pointer-events to avoid click, but it may still need an onsubmit prevent default on the form.

Beyond all this, if the form is submitted to runtest and there aren't enough runs, we should use our standard error template to note that.


deathbearbrown:
//TODO
- update userDetails query to contain the remainingRuns and monthlyRuns
- set remaining and monthly runs on the user (it's in the braintreeCustomerDetails query)
-  add conditionals to show this messaging

deathbearbrown:
this is deployed here https://wptqa.webpagetest.org/

deathbearbrown (approved):

PR #2083, opened 2022-06-28
Add header to force har to download

vibaldem:
Download HAR option loads contents into browser instead of downloading into a har file

### DIFF
--- a/www/export.php
+++ b/www/export.php
@@ -42,6 +42,7 @@ if (!strlen($filename))
     $filename = "pagetest";
 $filename .= ".$id.har";
 header('Content-type: application/json');
+header("Content-disposition: attachment; filename=$filename");
 
 // see if we need to wrap it in a JSONP callback
 if( isset($_REQUEST['callback']) && strlen($_REQUEST['callback']) )


tkadlec:
Nice catch! Instead of adding the header though, I'm thinking we should use the `download` attribute (https://caniuse.com/download) on the link itself, like we do for the JSON file here: https://github.com/WPO-Foundation/webpagetest/blob/7c69cbd99aa361bb415f60d6d2fa912a3ac4b781/www/testinfo_command-bar.inc#L21

Should probably just add that to all those download links.... 🤔

vibaldem:
> Nice catch! Instead of adding the header though, I'm thinking we should use the `download` attribute (https://caniuse.com/download) on the link itself, like we do for the JSON file here:
> 
> https://github.com/WPO-Foundation/webpagetest/blob/7c69cbd99aa361bb415f60d6d2fa912a3ac4b781/www/testinfo_command-bar.inc#L21
> 
> Should probably just add that to all those download links.... 🤔

That's neat, I didn't know about that attribute.  Only issue seems to be that it saves it as a .json instead of .har, at least it does on our private instance.

tkadlec:
@vibaldem Yeah...good point. So we need this after all. :) 

Ok, merging this. Thanks for jumping on it!

PR #2084, opened 2022-06-28
Update state on forms

deathbearbrown:
What: We needed Selects for choosing US state when signing up or modifying any addresses in the site
Every other country should still be able to manually enter text.

To test there's 3 areas where this is used:

- Signing up for a pro account through the sign up page

- signing up for a pro account through the account page

- updating your payment info on the account page

The address form defaults to United States and showing the state dropdown. If you change countries it will convert the state dropdown to an input field. If you change it back to US, it will return to being a dropdown of all the states.


tkadlec (approved):
This is looking solid to me, @deathbearbrown!

deathbearbrown:
tested on QA for all the areas that you can change your address.  Submitted the forms too, everything is working good, going to merge

PR #2088, opened 2022-06-30
Add remaining runs backend 

deathbearbrown:
This is a PR that merges the backend for remaining Runs into Scott's branch with the UI for it

### DIFF
--- a/www/index.php
+++ b/www/index.php
@@ -265,8 +265,8 @@ $remaining_runs =  (isset($request_context) && !is_null($request_context->getUse
                                                 </div>
 
                                                 <div class="test_presets_easy_submit">
-                                                    <small class="test_runs"><span><?= $remaining_runs; ?> Runs Left</span> | <a href="/account">Upgrade</a></small>
-                                                    <input type="submit" name="submit" value="Start Test &#8594;" class="start_test">
+                                                    <small class="test_runs <?= $remaining_runs === 0 ? 'runs-warn' : ''; ?>"><span><?= $remaining_runs; ?> Runs Left</span> | <a href="/account">Upgrade</a></small>
+                                                    <input type="submit" name="submit" value="Start Test &#8594;" class="start_test" aria-disabled="<?= $remaining_runs == 0 ? true : false; ?>">
                                                 </div>
                                             </div>
                                         </li>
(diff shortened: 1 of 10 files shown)


www/index.php
```diff
@@ -78,395 +78,396 @@
...
+                                                            <label for="private-simple"><input type="checkbox" name="private" id="private-simple" class="checkbox"> Make Test Private <small>Private tests are only visible to your account</small></label>
+                                                        </div>
+                                                    <?php endif; ?>
+                                                </div>
+
+                                                <div class="test_presets_easy_submit">
+                                                    <?php if ($is_logged_in) : ?>
+                                                        <small class="test_runs <?= $remaining_runs === 0 ? 'runs-warn' : ''; ?>"><span><?= $remaining_runs; ?> Runs Left</span> | <a href="/account">Upgrade</a></small>
+                                                    <?php endif; ?>
+                                                    <input type="submit" name="submit" value="Start Test &#8594;" class="start_test" aria-disabled="<?= $remaining_runs == 0 ? true : false; ?>">
```
jefflembeck, line 274: I'd make this `remaining_runs <= 0`,  because distributed systems and eventual consistency. There's a chance we let people run past their allotted amount ¯\_(ツ)_/¯

jefflembeck: also, do we want to add a `disabled=disabled` attribute if the same condition is met?

www/index.php
```diff
@@ -78,395 +78,396 @@
...
-                            </select>
-                            <span class="pending_tests hidden" id="pending_tests"><span id="backlog">0</span> Pending Tests</span>
-                            <span class="cleared"></span>
-                        </div>
+                                            </div>
+                                            <div>
+                                                <?php if ($is_logged_in) : ?>
+                                                    <small class="test_runs <?= $remaining_runs === 0 ? 'runs-warn' : ''; ?>"><span><?= $remaining_runs; ?> Runs Left</span> | <a href="/account">Upgrade</a></small>
+                                                <?php endif; ?>
+                                                <input type="submit" name="submit" value="Start Test &#8594;" class="start_test" aria-disabled="<?= $remaining_runs == 0 ? true : false; ?>">
```
jefflembeck, line 368: See 🔴

www/index.php
deathbearbrown: yes! I just noticed that aria-disabled is not cutting it when I QA'd it

www/runtest.php
```diff
@@ -2,920 +2,925 @@
...
+$error = NULL;
+$xml = false;
+$usingAPI = false;
+$usingApi2 = false;
+$forceValidate = false;
+$runcount = 0;
+$apiKey = null;
+$is_mobile = false;
+$isPaid = !is_null($request_context->getUser()) && $request_context->getUser()->isPaid();
+$hasRunsAvailable = !is_null($request_context->getUser()) && $request_context->getUser()->hasEnoughRemainingRuns();
```
jefflembeck, line 65: Unless we pass in the total runs being run, this will test if the person has more than 1 run.

deathbearbrown:
Screen Shot 2022-07-06 at 12 13 55 PM

www/templates/account/my-account.php
```diff
@@ -1,65 +1,82 @@
...
-  </div>
 
-<?php if ($is_paid) {
-  include_once __DIR__ . '/includes/billing-data.php';
-  include_once __DIR__ . '/includes/modals/subscription-plan.php';
-  include_once __DIR__ . '/includes/modals/payment-info.php';
-} else {
-  include_once __DIR__ . '/includes/signup.php';
-} ?>
+    <?php if (!$is_paid) : ?>
```
jefflembeck, line 58: Do we want to move this block down to L77 below?

PR #2106, opened 2022-07-07
Added ADR #3 re: jQuery

stoyan:


docs/adr/0003-remove-jquery-dependency.md
```diff
@@ -0,0 +1,20 @@
...
+- It's currently responsible for the 80%+ (LOC) in `site.js`
+
+
+## 🤔 Decision
+Remove jQuery and jQuery UI dependencies. Dialogs and tabs can be implemented in CSS with a sprinkle JavaScript.
+
+## 🎬 Consequences
+- Remove XSS vulnerability
+- Lighter page
+
```
jefflembeck, line 18: Is the full list of dependencies _just_ dialogs and tabs right now?

docs/adr/0003-remove-jquery-dependency.md
```diff
@@ -0,0 +1,20 @@
...
+🙋🏽‍♀️ **Status** Proposed
+
+## ℹ️ Context
+- The current use of jQuery 1.7.1 is a low vulnerability (XSS)
+- There is not much use of jQuery and jQuery UI, mainly dialogs and tabs
+- It's currently responsible for the 80%+ (LOC) in `site.js`
+
+
+## 🤔 Decision
+Remove jQuery and jQuery UI dependencies. Dialogs and tabs can be implemented in CSS with a sprinkle JavaScript.
```
tkadlec, line 13: Feel like we can ditch the "dialogs and tabs" bit and expand a bit.

Something like....

"We can replace existing jQuery dependencies, and build new components, by focusing on a strong foundation of markup and CSS, with JavaScript sprinkled in as necessary for enhancement."

tkadlec: Also, maybe we should make it explicit for anyone later that we intend on using vanilla JS at the moment. 

Again...

"We will be using vanilla JavaScript, rather than immediately switching to another framework, to keep the performance overhead to a minimum."

docs/adr/0003-remove-jquery-dependency.md
deathbearbrown: I think there's some stuff in www/site.js
I saw a `$.ajax`

deathbearbrown: Here's a list of what I see in there:
- drag and resizing (not sure where this is used)
- viewport offsets
- getting OS scrollbar width
- displaying local time
- tooltips
- scroll handling 
- editting test labels

jefflembeck (approved):
Heck yes. Looks like this could be broken down into tickets and is a pretty easily defineable project when we're ready

PR #2120, opened 2022-07-12
Avoid four 301 redirects

stoyan:
https://www.webpagetest.org/result/220712_BiDc1W_FQ/1/details/ shows 4 redirects from `webpagetest.org` to `www.webpagetest.org`


tkadlec (approved):
Woof. Good catch!

PR #2138, opened 2022-07-13
The Bones of a design system

deathbearbrown:
Hi I ripped out some design system bones from the account redesign just to show what it looks like with the css variables.

This is a PR for discussing how we'd like to do things. 

GOALS:
- discussing
- come up with some standards in terms of how we want to write css going forward
- naming conventions

I will write up another ADR after this with any important decisions we make so we have our coding standards documented somewhere. :)


www/css/button.css
```diff
@@ -31,30 +33,118 @@
...
-  border: 1px solid rgb(204, 204, 204)
+  border: var(--grey-outline)
+}
+
+/**
+ * ==============================================
+ * Dot Pulse
+ * ==============================================
+ */
+.dot-pulse {
```
deathbearbrown, line 69: this is an animation I'm adding to buttons when a form is submitted so people know to chill...

www/css/colors.css
```diff
@@ -0,0 +1,9 @@
+:root {
+  --body-text-black: #111;
+  --brand-yellow: #F9D856;
+  --brand-blue: #3881E5;
+  --brand-green: #80BC06;
+  --brand-red: #d02424;
+  --brand-purple: #6b25cf;
+  --grey-outline: 1px solid rgb(204, 204, 204);
```
deathbearbrown, line 8: I could call this colors or this could be brand, or it could just be variables.css.  Thoughts?

www/css/typography.css
```diff
@@ -0,0 +1,113 @@
+@import 'colors.css';
+
+body {
```
deathbearbrown, line 3: I have not included this file in the app in this PR (it is included in the account redesign) but here it is to show kind of the gist of what I'm trying to pull out and make consistent

deathbearbrown:
Also this is for @scottjehl, do you want to do BEM?

www/css/button.css
tkadlec: ❤️ this

deathbearbrown:
gonna close this and bring up bem at the next meeting :)

PR #2159, opened 2022-07-15
this fixes #2132 for basic auth specifically, but is not yet optimized. This logic could stand to be central and shared

### LINKED ISSUE #2132: Allow for some if not all Custom Scripting to apply alongside experiments
Currently, experiment scripts run in place of any scripting that's applied to a test. Many scripting features should be compatible with experiments, while some (like multistep tests) may not. 

We should aim to preserve scripting in experiments.

scottjehl:
Note: these changes will still not preserve a test['script'] field at all. They simply re-apply customHeaders and auth into the new script we're creating for an experiment. https://github.com/WPO-Foundation/webpagetest/issues/2132

### DIFF
--- a/www/runtest.php
+++ b/www/runtest.php
@@ -925,6 +925,16 @@ else {
                             $recipeScript .= ";";
                         }
                     }
+                    
+                    // Handle HTTP Basic Auth
+                    // TODO centralize this logic as it's borrowed from above temporarily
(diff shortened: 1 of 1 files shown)


scottjehl:
Tested and confirmed for auth and custom headers! Will follow up with Issue re cleanup and one small issue surfacing auth info in the compare page

PR #2173, opened 2022-07-18
target="_blank"-ify all tips

stoyan:
Some of the tips are missing the target and we do promise that all links open in a new tab.

Also took the opportunity to format the HTML.


tkadlec (approved):
:shipit:

PR #2174, opened 2022-07-19
Testlog refactoring: move single user history to a separate template

stoyan:
Tested webpagetest.org/testlog/7/ - looks identical before/after the change

### DIFF
--- a/www/templates/testhistory/user.php
+++ b/www/templates/testhistory/user.php
@@ -1,0 +1,90 @@
+<div class="history_hed">
+    <h1>Test History</h1>
+
+    <form name="filterLog" method="get" action="/testlog.php">
+
+        <?php if (!$is_logged_in) : ?>
+            <div class="logged-out-history">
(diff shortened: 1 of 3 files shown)


www/templates/testhistory/user.php
```diff
@@ -0,0 +1,90 @@
...
+                <a href="<?= "{$protocol}://{$host}/signup" ?>" class="btn-primary">Get Free Access</a>
+            </div>
+        <?php endif; ?>
+        <div class="history_filter">
+            <label for="filter">Filter test history:</label>
+            <input id="filter" name="filter" type="text" onkeyup="filterHistory()" placeholder="Search">
+            <?php if ($is_logged_in) : ?>
+                <label for="days" class="a11y-hidden">Select how far back you want to see</label>
+                <select name="days">
+                    <option value="1" <?php if ($days == 1) {
```
deathbearbrown, line 18: I see this formatting and I think "I'm gonna end up using a lot of ternary operators to not see this formatting"

deathbearbrown (approved):

Issue #1720, opened 2022-02-03
Colour contrast for links in Advanced Configuration


fershad:
There are three hyperlinks in the Advanced Configuration section that are dark blue (#1071ba) on slightly darker blue (#24334F). They're very hard to read.

The links are:
- One in the SPOF tab (text: [blackhole.webpagetest.org](https://blog.patrickmeenan.com/2011/10/testing-for-frontend-spof.html))
- Two in the Chromium tab:
  - Text: [V8 Runtime Call Stats](https://v8.dev/docs/rcs)
  - Text: [Host Resolver Rules](https://peter.sh/experiments/chromium-command-line-switches/#host-resolver-rules)

![image](https://user-images.githubusercontent.com/27988517/152352267-01f2f1fe-f27e-4a3c-bf12-0d15ec2944d3.png)

It would be nice if a lighter colour could be used.

PR #2176, opened 2022-07-20
White links on the homepage

stoyan:
Fix stuff like:

Screen Shot 2022-07-20 at 7 20 33 PM
Screen Shot 2022-07-20 at 7 20 59 PM

Looks like we have a number of `color:#fff` and `color:inherit` overrides, but I didn't touch these.

### DIFF
--- a/www/pagestyle2.css
+++ b/www/pagestyle2.css
@@ -48,7 +48,11 @@ select {
   color: #fff;
 }
 
-/* 
+.home a {
+  color: #fff;
+}
+
+/*
 header {
     grid-column-gap: 1.25em;
     border-bottom: 5px solid #1E2B47;
@@ -1858,7 +1862,7 @@ footer ul {
 
 /* Header */
 
-/* 
+/*
 #wptAuthBar {
     list-style-type: none;
 }
@@ -2234,7 +2238,7 @@ footer ul {
   /* margin-right: .5em; */
 }
 
-/* 
+/*
 .result-summary .opportunities_summary {
     display: block;
     max-width: auto;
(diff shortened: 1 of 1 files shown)


tkadlec:
Oooo… @stoyan, does this close #1720 ?

www/pagestyle2.css
```diff
@@ -48,7 +48,11 @@ select {
   color: #fff;
 }
 
-/* 
+.home a {
```
deathbearbrown, line 51: this will nuke those bottom links in the white boxes
Screen Shot 2022-07-26 at 1 07 17 PM

stoyan: Good catch, thanks! I didn't see these in my local setup. 

Do you think these should be the exception (and keep the general "all links are white on the homepage"), since all the rest of the page is dark background? Or I should drop the generic `.home a` and fix the offending links in the other parts?

stoyan:
> Oooo… @stoyan, does this close #1720 ?

yup

stoyan:
White links in the configuration area, sans the notification messages.
Screen Shot 2022-07-27 at 11 05 32 AM
Screen Shot 2022-07-27 at 12 09 01 PM
Screen Shot 2022-07-27 at 12 09 09 PM
Screen Shot 2022-07-27 at 12 09 31 PM

www/pagestyle2.css
deathbearbrown: I looked around and I couldn't find anything else weird. just those boxes which you can fix by applying a style to the box links

deathbearbrown (approved):

stoyan:
Yes, it does

PR #2188, opened 2022-07-27
bugfix: set the billing cycle select to annual.  Webkit is not our friend

deathbearbrown:
To test:
on signup select a MONTHLY plan

use the browser's back button and you should see the billing cycle select set to ANNUAL, not monthly, Monthly was the bug


jefflembeck (approved):
It'd be nice to be able to take advantage of this feature, but just making all of them conform is good enough.

deathbearbrown:
> 

I wasn't even getting value shown back... so I have no idea how we'd take advantage of it

PR #2205, opened 2022-07-29
Bring Laravel's Blade templates

stoyan:
1. Dependencies. Illuminate v9.0 libraries require PHP 8.2 so sticking with 8.0 for now
2. `www/resources/view.php` is the plumbing to add `view($tpl, $vars)`
3. couple of components - `x-fonts` and `x-favicons`


jefflembeck (approved):
This is fantastic. I can't wait to see this used.

PR #2207, opened 2022-08-01
Remove lib/json.php dependency

pmeenan:
All relatively recent versions of the built-in JSON support work well and the json.lib was used for pre-5.3 php (which is long-since not supported).  It also doesn't work with php 8 so this removes the dependency and cleans up any references to it.

From a user-facing perspective, this fixes HAR export on php 8.1 where the library was imported regardless of the php version in export.php (but not actually used there).


jefflembeck (approved):
Awesome. Thanks!

PR #2208, opened 2022-08-01
Removed the last json lib dependency

pmeenan:
Sorry, forgot to save the change to common.inc before submitting the last PR.  This removes the last dependency on the json lib (it's dynamic and won't normally trigger).


pmeenan:
Sorry, looks like some extra commits for script injection snuck in there - submitting a clean one in a second.

PR #2217, opened 2022-08-02
Make render-blocking requests for css and js more reliable in browsers that do not provide that information

scottjehl:
This adds and utilizes a new custom metric that collects scripts and links in the page that have render-blocking qualities.

This list is not a guarantee to contain render-blockers, as these may have been added dynamically by the time this runs. Still, we can use this list to compare in browsers that don't send a render-blocking flag on requests, so we can be more confident that a request that arrived before start render was actually render-blocking. Browsers such as chromium currently give us that information directly so we do not need this in those browsers.


stoyan:
nit: could you run `composer format:prettier`? I think it will have some opinions of quotes and such

tkadlec (approved):
Think this one is looking pretty good, @scottjehl!

PR #2229, opened 2022-08-02
dev.php for developer settings

stoyan:
So we can run a tight ship fixing warning without affecting prod.

Step to use:
1. rename `dev.sample.php` to `dev.php`
2. tweak to taste
3. fix everything
4. profit


jefflembeck (approved):
I dig this. Nice way to add dev-environment based instruction.

PR #2237, opened 2022-08-02
fix tag-label design

anthonyandersonfg:
The tag label does not display the text properly if the text has more than 21 characters:

image

To fix the problem I added the _white-space: nowrap;_ to the tag class:

image

But even though the text is larger than the container, the ellipsis is not being displayed:

image

That's because the _text-overflow: ellipsis;_ is not working properly for this tag, because it is assigned as a flex element, and [text-overflow works for block elements only](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow#formal_definition)
To fix I have added the _display: inline-block !important_.

That's how the tag will look when a text overflow occurs:  

image

### DIFF
--- a/www/pagestyle2.css
+++ b/www/pagestyle2.css
@@ -4186,10 +4186,12 @@ ul.meta_exp_applied_list {
 }
 
 span.test_presets_tag-label {
+  display: inline-block !important;
   background: #fce075 !important;
   color: #333333 !important;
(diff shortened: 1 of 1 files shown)


scottjehl:
This is a great tip. We can sort out our own use of !important here that this is repeating in a later cleanup. Thanks!

PR #2239, opened 2022-08-03
Hello Stylelint

stoyan:
1. Add `composer lint:css`
2. Use the lint config from https://github.com/stylelint/stylelint-config-recommended excluding `no-descending-specificity`
3. Fix one stray comment in `pagestyles2.css` that was tripping the parser

I excluded https://stylelint.io/user-guide/rules/list/no-descending-specificity/ because it makes the linter chokes on `pagestyles2.css` (I didn't realize this file is 10K lines long!) and it also finds too many things to fix. I'm thinking we can start fixing with the current config of rules, then add `no-descending-specificity`, then "upgrade" to https://github.com/stylelint/stylelint-config-standard

Most of the current problems are of the duplicates variety, either duplicate properties within the same block or duplicate selectors sprinkled in the same file. And a bunch of empty blocks too.

Currently `composer lint:css` may have overwhelming output, so probably a good idea is to start with individual files like:
`$ npx stylelint@14.9.1 www/wpt-header.css`

### DIFF
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -1,0 +1,2 @@
+www/chrome
+www/css/vendor
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,0 +1,44 @@
+{
+  "rules": {
+    "at-rule-no-unknown": true,
+    "block-no-empty": true,
+    "color-no-invalid-hex": true,
+    "comment-no-empty": true,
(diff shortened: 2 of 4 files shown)


jefflembeck:
This works as a good first step, for me.

jefflembeck (approved):

PR #2256, opened 2022-08-09
Change how we do default plan

deathbearbrown:
rendering "checked" on the radiobuttons is causing a WEIRD BUG. so lets do it in JS

### DIFF
--- a/www/templates/account/includes/signup.php
+++ b/www/templates/account/includes/signup.php
@@ -240,7 +240,7 @@
             foreach ($monthly_plans as $key => $plan) :
                 $plan_block = <<<HTML
                   <div class="form-wrapper-radio">
-                    <input type="radio" id="monthly-{$plan['id']}" name="plan" value="{$plan['id']}" r data-cost="{$plan['price']}" data-cycle="{$renewalDateMonthly}" equired />
+                    <input type="radio" id="monthly-{$plan['id']}" name="plan" value="{$plan['id']}" data-cost="{$plan['price']}" data-cycle="{$renewalDateMonthly}" required />
                     <label class="card wpt-plan" for="monthly-{$plan['id']}">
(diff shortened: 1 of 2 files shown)


tkadlec (approved):

www/templates/account/includes/signup.php
```diff
@@ -235,17 +235,19 @@
...
-        <input type="radio" id="monthly-{$plan['id']}" name="plan" value="{$plan['id']}" required />
-        <label class="card wpt-plan" for="monthly-{$plan['id']}">
-          <h5>{$plan['name']}</h5>
-          <div><strong>\${$plan['price']}</strong>/Month</div>
-          <span aria-hidden="true" class="pill-button yellow">Select</span>
-        </label>
-      </div>
-HTML;
+                  <div class="form-wrapper-radio">
+                    <input type="radio" id="monthly-{$plan['id']}" name="plan" value="{$plan['id']}" r data-cost="{$plan['price']}" data-cycle="{$renewalDateMonthly}" equired />
```
jefflembeck, line 243:
```suggestion
                    <input type="radio" id="monthly-{$plan['id']}" name="plan" value="{$plan['id']}" data-cost="{$plan['price']}" data-cycle="{$renewalDateMonthly}" required />
```

www/templates/account/includes/signup.php
```diff
@@ -329,6 +336,18 @@
 <script src="https://js.braintreegateway.com/web/dropin/1.33.0/js/dropin.min.js"></script>
 
 <script>
+const radios = document.querySelectorAll('input[name="plan"]')
```
jefflembeck, line 339: Do we need to add a listener for each radio button? Wondering if a single listener at a parent element might be easier?

maybe that's too much and we can worry about it later

jefflembeck (approved):

PR #2258, opened 2022-08-09
2257 - clarify right col for account steps 2 and 3

scottjehl:


www/templates/account/signup/step-2.php
```diff
@@ -46,28 +46,32 @@
...
-                <?php endif; ?>
-                <th>Price</th>
-            </thead>
-            <tbody>
-                <tr>
-                    <?php if ($is_plan_free) : ?>
-                        <td>Free</td>
+            <?php if( !$is_plan_free ) : ?>
+            <tr>
+                <th>Pay Plan:</th>
```
deathbearbrown, line 56: This whole sidebar should be an include, so we don't repeat it OOOH NOOO I AM DRYing

deathbearbrown (approved):
@jefflembeck for the trumpet branch, lets make this area an include

Issue #2277, opened 2022-08-17
Remove Pricing from Nav for people running their own instances


deathbearbrown:
## Description
Pricing goes to the sign up page and should not be present for people who are running the app in their own instances
![Screen Shot 2022-08-17 at 9 12 09 AM](https://user-images.githubusercontent.com/607541/185189835-2438fdaa-81a8-43e7-9b38-06a71378f086.png)

### Steps to reproduce
Run your own instance of WPT, See Pricing in the nav

### Expected result
Pricing should not be there, just as sign up and login isn't present.

### Actual result
Pricing link in nav goes to sign up page which is broken

PR #2279, opened 2022-08-17
don't show pricing for private instances

deathbearbrown:
#2277 
Remove Pricing from the nav for private instances. (big thanks to @sammeboy635 for his work on the docker instance which is how we found the bug!)


stoyan (approved):
shipit

jefflembeck (approved):

PR #2280, opened 2022-08-17
ADR 4: Migrate to Laravel

stoyan:


docs/adr/0004-adopt-laravel.md
```diff
@@ -0,0 +1,106 @@
...
+Adopt [Laravel](https://laravel.com/) for the www codebase. Laravel can handle API requests, cron jobs, command line (console) tools, but we'll limit this ADR to web only. Follow up ADRs will deal with additional Laravel adoption.
+
+
+## 🎬 Consequences
+Migrating the home-grown codebase to a well-documented popular framework will help:
+- people orient easier (and find answers to common questions on stackoverflow) and make a change in confidence.
+- hire engineers with potentially some Laravel experience
+
+## Process
+We want the migration to be gradual and as obtrusive as possible regarding day-to-day coding. We can attack the pieces (Routing, Models, Views, Controllers) separately and asynchronously.
```
jefflembeck, line 21:
```suggestion
We want the migration to be gradual and as unobtrusive as possible regarding day-to-day coding. We can attack the pieces (Routing, Models, Views, Controllers) separately and asynchronously.
```

jefflembeck (approved):
Other than the typo nitpick, this seems like a great approach. Dumping routing from nginx.conf so we can have a single point of entry will certainly be some lifting, but feels like a totally doable first big move.

PR #2286, opened 2022-08-19
Moved all js, css, images under www/assets/ and updated references

stoyan:
In preparation to the [Laravel migration](https://github.com/WPO-Foundation/webpagetest/blob/master/docs/adr/0004-adopt-laravel.md) we need to cleanup `www/`

Sanity testing by randomly browsing the site and results pages.


deathbearbrown (approved):
@jefflembeck will this be a bananas rebase into cue-trumpet?

jefflembeck:
@deathbearbrown not total bananas. Like one banana. Maybe two.

PR #2299, opened 2022-08-24
Allow file uploads in the Script tab

stoyan:
Demo, using the new recorder extension:

https://user-images.githubusercontent.com/51308/186512589-050d7ee8-4edd-4203-9959-5f9e8ebd1795.mov

### DIFF
--- a/www/js/site.js
+++ b/www/js/site.js
@@ -1193,3 +1193,26 @@ var getScrollbarWidth = function() {
     };
   }
 })();
+
+// handle script file uploads
+const uploadInput = document.getElementById("script_file");
+if (uploadInput) {
+  uploadInput.addEventListener(
+    "change",
+    () => {
+      const file = uploadInput.files[0];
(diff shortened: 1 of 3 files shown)


www/js/site.js
```diff
@@ -1193,3 +1193,26 @@ var getScrollbarWidth = function() {
...
+if (uploadInput) {
+  uploadInput.addEventListener(
+    "change",
+    () => {
+      const file = uploadInput.files[0];
+      const reader = new FileReader();
+      reader.addEventListener(
+        "load",
+        () => {
+          document.getElementById("enter_script").value = reader.result;
```
deathbearbrown, line 1208: How do we prevent this from being used maliciously?

deathbearbrown (approved):

www/js/site.js
stoyan: Thanks! As discussed offline, we should be fine (famous last words) :)

PR #2305, opened 2022-08-30
Yank jquery out of site.js and minify the unminified parts

stoyan:
The file `jquery.js` is loaded unversioned with the idea that we'll never touch it other than to delete it. But we can always version it later, if necessary.

For testing, I browsed (with console open) the home page and results page, all seems to work fine.

The page https://www.webpagetest.org/webvitals.php though seems to be loading all scripts twice, in prod too so unrelated to this change


jefflembeck (approved):
🙌 🐑 it

PR #2306, opened 2022-08-30
Deploy to Prod

tkadlec:
Got some nice stuff, let's get it out there. :)


stoyan:
Note that things not in this repo might be affected by the assets move. The only one I can think of is the twitch logo from settings.ini

tkadlec:
@stoyan Gotcha. That's covered already from the looks of it, so this feels pretty safe.

PR #2311, opened 2022-08-31
Enterprise user account page + Sessions + Cleanup

jefflembeck:
This PR does a few things (sorry).

- Destroys sessions on logout. I don't need to keep around php sessions for users who are logging out. We have no need for that information
- Remove billing information for enterprise user account pages. We're having issues with getting any data that isn't api keys and personal info right now and will work on making the other data come back
- Format and clean up some code
- Add a little debug thing for when a user is in dev mode


deathbearbrown:
do you have this deployed anywhere?

deathbearbrown (approved):

PR #2315, opened 2022-09-01
Making room for index.php to do routing

stoyan:
A step in the direction of the [routing part of the Laravel transition](https://github.com/WPO-Foundation/webpagetest/blob/master/docs/adr/0004-adopt-laravel.md#routing)


deathbearbrown:
do you also need to update nginx's entry point to home.php?

stoyan:
nah, the entry point is still index.php, home.php is just being included

PR #2318, opened 2022-09-01
feat(user): add first/last/company name

jefflembeck:
you can have a little $user->getFirstName() and $user->getLastName(),
and $user->getCompanyName() [as well as all of their sets] as a treat.

This is a multi-piece job.

- we add first, last, and company name all to the User object.
- we query for those things on the activeContact on every page load
    - This allows us to take out the getUserContactInfo query,
      making the account page faster
- we move all of that setting of things out of the middleware and into
  the client itself and have the client return a user object. This is so
  we have better tooling and we stop doing _so_ many things in two
  places.
- add some tests to make sure this all works

### DIFF
--- a/tests/CPClientTest.php
+++ b/tests/CPClientTest.php
@@ -261,17 +261,23 @@ final class CPClientTest extends TestCase
         $client->revokeToken($token);
     }
 
-    public function testGetUserDetails(): void
+    public function testGetUser(): void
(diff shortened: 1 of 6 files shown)


scottjehl:
niiiice

deathbearbrown (approved):

Issue #2308, opened 2022-08-30
webpagetest.org/webvitals has a JS error "sel" undefined


stoyan:
looks like we're looking for a select box that isn't there

to test go to webpagetest.org/webvitals with console open

PR #2328, opened 2022-09-07
Webvitals: unbreak the profile selection

stoyan:
s/profile/webvital_profile/

Results!
 * the JS error (`sel` is null so cannot read `sel.options`) is gone
 * the cookie `wvProfile` is written
 * the description of the profile shows up:
Screen Shot 2022-09-07 at 12 04 45 PM


stoyan:
Fixes #2308

PR #2331, opened 2022-09-07
Offer ability to remove TTFB variance from experiment comparison

### LINKED ISSUE #2329: Experiment Compare values could adjust for TTFB variance, at least optionally
Often, there's network variance between experiment and control runs that shows up in TTFB. Being additive, the TTFB diff can end up rippling into the diffs of later metrics that can have much smaller diffs if it weren't for the TTFB variance.

One approach to fixing this will be the longstanding idea of [normalizing TTFB between compare runs](https://github.com/WPO-Foundation/webpagetest/issues/1435), which we should still aim to offer. 

But another, and perhaps additional thing we can do in Experiment results would be to detect TTFB variance and offer a way to view the experiment results with TTFB removed from  chronological metrics before comparing them. That'd mean we could more reliably see diffs in the metric changes that we offer on that page.

here's an experiment with ttfb variance: https://www.webpagetest.org/video/compare.php?tests=220906_AiDcYK_EPH,220906_AiDcX7_EPG

and its results: 
image

Here's those same results adjusted for TTFB:
image

Offering this via query string and a link, non-default, would be a nice way to start. A message that says:

> We detected some variance in response time between your experiment and control runs, which might make other metric changes seem more significant than they really are. [View this comparison with TTFB differences removed](#)

scottjehl:
This implements a check for TTFB differences and when they are greater than 100ms, offers a link to view the comparison with TTFB ignored from all additive metrics. fixes #2329

### DIFF
--- a/www/experiments/findings.inc
+++ b/www/experiments/findings.inc
@@ -85,7 +85,7 @@ if ($experiment) {
                             $controlMetric = $controlMetric - $controlTTFB;
                         }
                         $thisDiff = $controlMetric - $experimentMetric;
-                        $units = "<span class=\"units\">" . $metric[2] . "</span>";
+                        $units = '<span class="units">' . $metric[2] . '</span>';
                         $changeTerms = array("Faster", "Slower");
                         if ($metric[3]) {
                             $changeTerms = $metric[3];
@@ -133,7 +133,7 @@ if ($ttfbDiffNotable) {
     if ($removeTTFBfromDiffs) {
         echo '<p class="ttfb_experiment_warning"><strong>Note:</strong> This comparison view is omitting TTFB to adjust for network variance. <a href="' . $experimentResultsHref . '">View this comparison with TTFB differences visible.</a>';
     } else {
-        echo '<p class="ttfb_experiment_warning"><strong>Note:</strong> We noticed a ' . $ttfbDiffClean . 's difference in response time between the experiment and control runs, which can make other metric changes seem more significant than they really are. <a href="' . $experimentResultsHref . '&ignoreTTFB">View this comparison with TTFB differences removed</a>';
+        echo '<p class="ttfb_experiment_warning"><strong>Note:</strong> This experiment had a ' . $ttfbDiffClean . 's difference in response time between the median experiment and control runs, which can make other metric changes seem more significant than they really are. To examine the variability of response times in all test runs, you can <a href="/graph_page_data.php?tests='.$tests[0]["id"]. ',' .$tests[1]['id'].'&medianMetric=ttfb&control=1">plot full results</a>. Or, you can <a href="' . $experimentResultsHref . '&ignoreTTFB">view this comparison with TTFB removed from later metric comparisons</a>.';
     }
 }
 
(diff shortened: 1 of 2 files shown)


www/experiments/findings.inc
```diff
@@ -23,86 +24,137 @@ if ($experiment) {
...
+    $removeTTFBfromDiffs = isset($_GET['ignoreTTFB']);
+
+
+    // check if metric might benefit from TTFB removal
+    function ignoreTTFBforMetric($metric)
+    {
+        global $removeTTFBfromDiffs;
+        if (
+            $removeTTFBfromDiffs
+            && ($metric === "render"
```
tkadlec, line 57: Instead of having to list out metric names again, could we append a value to the $diffMetrics array indicating whether or not a metric should be impacted by TTFB removal? Just thinking it avoids duplicating metric names.

scottjehl: sure! good call.

stoyan: yeah, we need to move all these strings as const/struct of some sort at some point. String are too easy to mistype

www/experiments/findings.inc
```diff
@@ -23,86 +24,137 @@ if ($experiment) {
...
+    $ttfbDiffClean = floatval(removeLeadingZero(round($ttfbDiff, 3)));
+    $ttfbDiffClean = abs(round($ttfbDiffClean / 1000, 2));
+    $ttfbDiffNotable = abs($ttfbDiff) > 100;
+    $removeTTFBfromDiffs = isset($_GET['ignoreTTFB']);
+
+
+    // check if metric might benefit from TTFB removal
+    function ignoreTTFBforMetric($metric)
+    {
+        global $removeTTFBfromDiffs;
```
stoyan, line 54: Why not pass this to the function instead of using a global?

www/experiments/findings.inc
```diff
@@ -23,86 +24,137 @@ if ($experiment) {
...
+    </div>
+
+
+<?php }
+
+if ($ttfbDiffNotable) {
+    if ($removeTTFBfromDiffs) {
+        echo "<p class=\"ttfb_experiment_warning\"><strong>Note:</strong> This comparison view is omitting TTFB to adjust for network variance. <a href=\"" . $experimentResultsHref . "\">View this comparison with TTFB differences visible.</a>";
+    } else {
+        echo "<p class=\"ttfb_experiment_warning\"><strong>Note:</strong> We noticed a " . $ttfbDiffClean . "s difference in response time between the experiment and control runs, which are usually unrelated to the experiment and can might make other metric changes seem more significant than they really are. <a href=\"" . $experimentResultsHref . "&ignoreTTFB\">View this comparison with TTFB differences removed</a>";
```
stoyan, line 156: I'd say single quotes here will be easier to read, since you're concatenating the php vars and not inlining them
```
echo '<p class="ttfb...
```

stoyan (commented):
This is awesome! Couple of nits inline...

www/experiments/findings.inc
scottjehl: done

stoyan (approved):

www/experiments/findings.inc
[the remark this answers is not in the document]
scottjehl: on it

PR #2338, opened 2022-09-09
Remove unused paths from nginx conf

stoyan:
These are currently 404s. Removed them all except https://webpagetest.org/events which is fixed. Although it should probably be just a redirect, but keeping it for now.


jefflembeck (approved):
🐑 it!

PR #2339, opened 2022-09-09
Remove support for legacy /results/ paths

stoyan:
These paths were introduced at least as back as 11 years ago from what I can tell:

https://github.com/WPO-Foundation/webpagetest/blame/3ff126c63761d0a44a22074a0b8ce3b994a94e23/www/nginx.conf

@pmeenan can you think of a reason to keep them?


pmeenan:
That actual /old/ ones are ancient and have long-since been deleted so zero risk with those.  I'm assuming the non-old waterfall and optimization check png's aren't used in the UI and the php is called directly but it would be worth verifying.

stoyan:
Ahaa, there's one leftover:
https://github.com/WPO-Foundation/webpagetest/blob/b4b99e59c90ddbe8e6b32f4ca85c7e8d0cf1bf80/www/include/UrlGenerator.php#L240-L249

Thanks for the quick response, @pmeenan!

stoyan:
Found a few more places, will address them separately and then come back to this

stoyan:
The holdouts are addressed in #2343

jefflembeck (approved):

Issue #2347, opened 2022-09-13
HTTP 103 Early Hints not reported correctly when multiple Link headers are used


paulcalvano:
It seems that HTTP Early hints are under-reported when multiple Link response headers are used.   When all the resource hints are included in a single Link response header, they are reported correctly.

For example, on [this](https://www-f-staging.etsy.com/?early-hints=single-link-header) page, the following Early hint is returned.

```
< HTTP/2 103 
< link: <https://www.etsy.com/>; rel=preconnect,<https://www.etsy.com/>; crossorigin; rel=preconnect,<https://i.etsystatic.com/>; crossorigin; rel=preconnect,<https://i.etsystatic.com/>; rel=preconnect,<https://js.sentry-cdn.com/>; rel=preconnect,<https://js.sentry-cdn.com/>; crossorigin; rel=preconnect,<https://www.googletagmanager.com/>; rel=preconnect,<https://www.googletagmanager.com/>; crossorigin; rel=preconnect
```

WebPageTest correctly shows the hints  ([Example test](https://www.webpagetest.org/result/220913_BiDc3V_DCA/1/details/#waterfall_view_step1))

<img width="1046" alt="image" src="https://user-images.githubusercontent.com/7459458/189985553-bd91add7-8fcf-4cff-9b63-3e87ccb45840.png">

On [this](https://www-f-staging.etsy.com/?early-hints=multiple-link-headers) page the resource hints are sent across multiple Link response headers, 

```
< HTTP/2 103 
< link: <https://www.etsy.com/>; rel=preconnect
< link: <https://www.etsy.com/>; crossorigin; rel=preconnect
< link: <https://i.etsystatic.com/>; crossorigin; rel=preconnect
< link: <https://i.etsystatic.com/>; rel=preconnect
< link: <https://js.sentry-cdn.com/>; rel=preconnect
< link: <https://js.sentry-cdn.com/>; crossorigin; rel=preconnect
< link: <https://www.googletagmanager.com/>; rel=preconnect
< link: <https://www.googletagmanager.com/>; crossorigin; rel=preconnect
```

I can see that the early hints were  successul in the waterfall graph, but they are under-reported in the request dialog  ([Example test](https://www.webpagetest.org/result/220913_BiDcWG_DCP/1/details/#waterfall_view_step1))

<img width="1110" alt="image" src="https://user-images.githubusercontent.com/7459458/189986143-9e6c9a24-005e-4834-a8c9-a898b760f197.png">

[...]

PR #2348, opened 2022-09-13
Handle multiple early hints headers

stoyan:
## Before:

Screen Shot 2022-09-13 at 5 51 25 PM

## After:

Screen Shot 2022-09-13 at 5 51 08 PM

Test: https://www.webpagetest.org/result/220913_BiDcWG_DCP/1/details/
Regression test: https://www.webpagetest.org/result/220913_BiDc3V_DCA/1/details/

Solves: #2347


tkadlec (approved):
Working great for both cases!

PR #2365, opened 2022-09-16
fix(user): TEMP FIX: enterprise users always have 1000 runs

jefflembeck:
We've got some issues with giving enterprise users real data on how many runs they have left. This oughta take care of that for now (kind of)


tkadlec (approved):
Neat and tidy! 🚢 it!

Issue #2373, opened 2022-09-23
Ability to pass current test ID into custom scripts


tkadlec:
## Feature Request Summary
The ability to use the current test ID in custom scripts with variable substitution (ex: %TESTID%)

### What's the motivation or use case for the change/feature?
We already let folks append the test ID to the ua string, but making it available in the custom scripts means they could do things like append the ID as a query parameter so that they can dig into the tracing/logs on their own architecture to follow the path of the test more clearly.

We currently do our substitution here: https://github.com/WPO-Foundation/webpagetest/blob/5378d7aecd9033174598640339d84ae37c804943/www/runtest.php#L3441

tkadlec:
closed by #2381

PR #2381, opened 2022-09-27
Ability to pass current %TEST_ID% into custom scripts

sammeboy635:
#2373 

Added %TEST_ID% to custom scripts
- I picked %TEST_ID over %TESTID% just to follow other variable substitutions like "%HOST_REGEX%"
- I can also add this to the WebPageTest-docs if you would like.

cbefore
cafter


stoyan:
Looks good, thanks!

Updating the docs would be much appreciated, thanks for offering.

PR #2383, opened 2022-09-27
fixes #2382.

### LINKED ISSUE #2382: Experiment Results are not displaying bytes downloaded differences in comparison
It's a typo in the reference, but also we'll need some math for the kb part

scottjehl:
Included is a slight font-size shift to make room for more metrics when needed.

Screenshots of KB addition:
image


stoyan (changes requested):
I think the diff switched from 2 spaces indentation to 4 in pagestlye2.css, so maybe try back to 2 and resend?

stoyan (approved):
shipit

PR #2398, opened 2022-10-03
Removed couple of unused constants

stoyan:
They are:
* always true
* not referenced in code


scottjehl:
This seems okay to me, but does the code have anything to do with private instance code?

jefflembeck (approved):
Not used is good for me :)

PR #2405, opened 2022-10-04
fix(customer): support expired payment methods

jefflembeck:
When a payment method expires in Chargify, they start sending back null. This causes errors for us


stoyan (approved):
shipit

PR #2410, opened 2022-10-04
Remove "Analyze JPEG" page

pmeenan:
The agents now collect a bunch of image information automatically at test time and include it with the test results (in the json).  There is no longer a need to analyze images directly server-side.

This removes the jpeginfo server-side analysis.

The image info is currently in the JSON data with each requests but can eventually be surfaced directly in the "view all images" UI as a better replacement.


jefflembeck (approved):
🐑  it!

tkadlec (approved):

PR #2412, opened 2022-10-04
Removed relay support

pmeenan:
The ability to relay tests through the public instance hasn't been supported for at least the last 2 years.  This cleans up the code to remove the relay support and should have caught all of the tendrils through the code.


jefflembeck (approved):
5 star PR right here. Would approve again.

PR #2414, opened 2022-10-04
Remove the test import functionality

pmeenan:
Removes the long-unsupported code for importing a new test based on Chrome dev tools captures (was originally for integration with CI systems that captured their own data but was never really used).


jefflembeck (approved):
YESSSS

PR #2415, opened 2022-10-04
Removed statsd, tsview integrations and auto APK updating

pmeenan:
This removes 2 unused integrations (sending data to a statsd back-end and a tsview back-end) as well as the logic that would auto-update Android APK's from a phone attached to the server.


pmeenan:
Having troubles getting rebase to work cleanly - I'll close this and re-do the PR

PR #2420, opened 2022-10-04
Fix 2406-  improve form usability in create experiments

### LINKED ISSUE #2406: Improve Create Experiment field usability
The create fields that offer free-form text input are hard to use. 

They should be textareas and have more room to type, with multiple lines if desired.

Current experience:
image

Proposed experience (toggle):
image

image

scottjehl:
fixes #2406 
screenshot:


stoyan (approved):
looks good

PR #2431, opened 2022-10-06
move constants together

stoyan:
The idea is to have a single place to look for all constants

* Found a couple (`BREAKDOWN_CACHE_VERSION` and `VIDEO_CODE_VERSION`) sprinkled in the codebase and moved them to the new `constants.inc`
* Deleted code around `ADULT_SITE` which is not referred to anywhere
* Moved `EMBED` to `constants.inc` and tweaked references because it's now always defined.
* Two constants `FRIENDLY_URLS` and `VER_TIMELINE` require `GetSetting()` from `common_lib.inc` so cannot be in `constants.inc`. Moved them closer to where constants.inc is required.

Tested by browsing the site with and without `?embed`, e.g.
https://www.webpagetest.org/video/compare.php?tests=220926_BiDcVY_E4Z-r:1-c:0-e:filmstrip&embed
vs
https://www.webpagetest.org/video/compare.php?tests=220926_BiDcVY_E4Z-r:1-c:0-e:filmstrip

### DIFF
--- a/tests/Handlers/AccountTest.php
+++ b/tests/Handlers/AccountTest.php
@@ -36,6 +36,7 @@ define('VER_ACCOUNT_CSS', "account.css");
 define('VER_JS_ACCOUNT', "account.js");
 define('VER_JS_ESTIMATE_TAXES', "estimate-taxes.js");
 define('FRIENDLY_URLS', true);
+define('EMBED', false);
 
 final class AccountTest extends TestCase
 {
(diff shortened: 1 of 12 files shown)


jefflembeck (approved):

jefflembeck:
what's the origin of the adult site part? is there a legal part to that? @pmeenan

pmeenan:
No, you can get rid of all of that.  It was from when the site used Adsense for ads, they didn't allow ads to be placed on pages with "adult content" so WPT had to jump through a bunch of hoops and exclude the ad code from those test results.

stoyan:
oh, that's good to know, thanks @pmeenan! I didn't dig deeper into some of the helper methods and keyword lists, but it's great we clean those up too. Task #2443

PR #2434, opened 2022-10-06
Autoload constants.inc

stoyan:
As part of the transition off `__DIR__` includes, make the `*_PATH` constants always available by adding them to autoload.

Ran `composer dump-autoload`, tested with `composer test` to make sure the `WWW_PATH` is available in tests

Browsed the site after removing the explicit require for constants.inc to make sure it's not broken and constants are all defined.

`composer dump-autoload` made some changes, hope they look ok.


jefflembeck (approved):
WOO. That's really nice.

PR #2435, opened 2022-10-06
Delete all $gaTemplate instances

stoyan:
`$gaTemplate` is defined all over the place but not used anywhere. Delete all instances so we don't keep copying it in new code


jefflembeck (approved):
WOO

PR #2447, opened 2022-10-07
Fixed undefined variable error in \www\page_data.inc

gardness:
Created a quick fix for this error.


stoyan (approved):
Excellent, thanks!

PR #2448, opened 2022-10-07
Fix __DIR__ use (in 30 out of 120-ish files)

stoyan:
Includes of common.inc and autoload.php need to remain with __DIR__ for now, as the constants are not yet loaded.

Will address the rest of the instances in follow up diffs, let's see what this one breaks.

### DIFF
--- a/www/consolelog.php
+++ b/www/consolelog.php
@@ -1,4 +1,5 @@
 <?php
+include 'common.inc';
 
 // this output buffer hack avoids a bit of circularity
 // on one hand we want the contents of header.inc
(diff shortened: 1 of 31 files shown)


www/consolelog.php
```diff
@@ -1,4 +1,5 @@
 <?php
+include 'common.inc';
```
jefflembeck, line 2: `require_once WWW_PATH . '/common.inc';`

Might be a little nicer here? Guarantees a single load, throws an error if there's no file found, and makes sure we can move it around with the new path! 😃

stoyan: the thing is autoload doesn't load `constants.inc` for the legacy code, so it will have to be __DIR__ when common.inc is included for now :(

stoyan: it's `require_once __DIR__ . '/common.inc';` now

PR #2451, opened 2022-10-10
A round of dead PHP code fixes

stoyan:
Ran a tool called PHPMD to look for unused variables
```
$ wget -c https://phpmd.org/static/latest/phpmd.phar
$ php phpmd.phar webpagetest/www ansi unusedcode --suffixes php,inc --exclude www/lib/,www/ec2/,www/resources/compiled
```
Looked around the output and partially applied changes where VSCode's PHP Intelephense seemed to agree with phpmd and changes looked safe, especially careful with the use of globals

### DIFF
--- a/www/video.inc
+++ b/www/video.inc
@@ -53,7 +53,7 @@ function BuildVideoScripts($testPath)
         while ($file = readdir($dir)) {
             $path = $testPath  . "/$file";
             if (is_dir($path) && !strncmp($file, 'video_', 6)) {
-                BuildVideoScript($testPath, $path);
+                BuildVideoScript($path);
             }
         }
 
@@ -66,15 +66,12 @@ function BuildVideoScripts($testPath)
 *
 * @param mixed $videoDir
 */
-function BuildVideoScript($testPath, $videoDir, $endTime = 0, $extend = false, $startOffset = null)
+function BuildVideoScript($videoDir)
 {
-  // figure out the frame range and actual file names
-    $start = null;
+    // figure out the frame range and actual file names
     $end = null;
     $frames = array();
-    if ($endTime) {
-        $endTime = (int)(($endTime + 99) / 100);
-    }
+    $endTime = 0;
 
     if (loadVideo($videoDir, $frames_ms) && count($frames_ms)) {
       // get 10fps frame timings for now - eventually add support for rendering 60fps video
--- a/www/video.inc
+++ b/www/video.inc
@@ -53,7 +53,7 @@ function BuildVideoScripts($testPath)
         while ($file = readdir($dir)) {
             $path = $testPath  . "/$file";
             if (is_dir($path) && !strncmp($file, 'video_', 6)) {
-                BuildVideoScript($path);
+                BuildVideoScript($testPath, $path);
             }
         }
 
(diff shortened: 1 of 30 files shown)


www/video.inc
```diff
@@ -66,15 +66,12 @@ function BuildVideoScripts($testPath)
 *
 * @param mixed $videoDir
 */
-function BuildVideoScript($testPath, $videoDir, $endTime = 0, $extend = false, $startOffset = null)
```
pmeenan, line 69: I'm pretty sure the whole "video scripts" thing is ancient and no longer used but removing it should probably be done separately.  AVI Synth hasn't been used for a long time and videos are rendered directly by the server now on-demand.

www/video/render.inc.php
```diff
@@ -757,26 +743,19 @@ function DrawTest(&$test, $renderInfo, $frameTime, $im)
         $path = $test['frames'][$frame_ms]['path'];
     }
 
-    $need_grey = false;
```
pmeenan, line 760: I'm a bit confused about the need_grey/applyGrey removal.

It looks like $applyGray defaults to true unless a query param explicitly disables it. The logic below determines if a given frame is the last frame and is past the end time of the video and if so, grays out the frame.

It is the logic for ending each side of a comparison video with a gray page content after the page is finished and doesn't appear to be unused.

www/waterfall.inc
```diff
@@ -1958,8 +1956,7 @@ function GetAverageSliceValue($slices, $start, $end)
             $start_value = floatval($slices[$start_key]);
             $end_value = floatval($slices[$end_key]);
             $end_weight = floatval($start - $start_bucket) / 0.1;
-            $start_weight = 1.0 - $distance_from_start;
-            $avg = $start_weight * $start_value + $end_weight * $end_value;
```
pmeenan, line 1962: The math here looks wrong.  It looks like $distance_from_start might not have been set and I can't remember exactly what it was supposed to be doing but it looks like it was supposed to be using part of each of the start and end values and removing the start_weight entirely doesn't fix the math (though probably makes it do what the code was actually doing - but not what it was intended to do).

pmeenan (changes requested):
Mostly looks good but I think the grayscale logic removal was incorrect and the math issue noted in the comments as well.  The video.inc video script logic can probably be removed entirely but that's probably better done as a separate PR.

Make sure to test it out on a dev server, particularly poking around at filmstrips and videos because that looks like where most of the logic was removed.

www/waterfall.inc
stoyan: yeah, `$distance_from_start` is never defined anywhere and so in PHP 1.0 - undefined var is 1, so `$start_value` times 1 is always `$start_value`. But I see what you mean about the logic. I'll just define `$distance_from_start = 0` and leave it alone

www/video.inc
stoyan: Reverted all `BuildVideoScript()` changes

www/video/render.inc.php
stoyan: so `$applyGrey` was defined in one function, used in another, so it wasn't doing anything which lead me to delete a bunch of things. Now I moved it to where it's used and the feature is back on (currently broken in www, see https://www.webpagetest.org/video/view.php?tests=221010_AiDcVV_EXQ-r:1-c:0)

pmeenan (approved):

pmeenan:
LGTM, thanks for the cleanup.

PR #2454, opened 2022-10-11
Restore LCP background image Opp & Exp

### LINKED ISSUE #2449: LCP opportunities are not finding background image LCPs now

scottjehl:
It seems that the LCP info coming from the browser may have changed recently and no longer includes bg image lcp info on the element object of the LCP event.

But the info is often elsewhere at least in the json.

This keeps our existing logic for old tests with test data that contains LCP bg image info on the element, while adding a check for the new location

fixes #2449


stoyan (approved):
shipit

PR #2459, opened 2022-10-12
Duplicate selectors cleanup in pagestyle2

stoyan:
I noticed some duplicate declarations, such as `.center{}` and `.left{}` so did a round of simplest duplicate selectors removal with the help of 

`npx stylelint@14.9.1 www/assets/css/pagestyle2.css` with the `no-duplicate-selectors` check added to `.stylelintrc`

There's definitely more cleanup to do, however it may require more testing, this is just the obvious stuff.


jefflembeck (approved):
🐑  it

Issue #2452, opened 2022-10-10
Scrollable tables rendering hiccups


stoyan:
Several big scrollable tables in the results have hiccups rendering and only show up on scrolling.

Example: https://www.webpagetest.org/result/221007_AiDc8P_FYA/1/details/

https://user-images.githubusercontent.com/51308/199577550-9083c0db-c524-4abd-b749-f05f5537b134.mov

Edit: URL and video to show `mx-blend-mode` on/off

stoyan:
Viewport size happen to be 2002x1146 (w x h)

hecktarzuli:
I'm also seeing some funky tables.  Macbook Pro 16" 2019, Chrome 105.  1920x1080.

https://youtu.be/3-E2raj55yk

stoyan:
Appears to happen only in Chrome (including Canary 109.0.5396.0) and Brave. It's fine in Safari and Firefox.

yoavweiss:
@stoyan - Could y'all file an issue on https://crbug.com with a reproducible test case?

PR #2461, opened 2022-10-12
Replaced glossary table with a definition list

stoyan:
Optimization glossary table looked a bit dated (smaller font) and glitchy to render (see #2452), so replaced it with a `<dl>`. Also deleted CSS for `table.glossary`, it wasn't used.

Any comments appreciated regarding content (didn't touch it) or formatting.

Before: 

(no rendering till scroll)

Screen Shot 2022-10-12 at 3 55 07 PM

(scroll)

Screen Shot 2022-10-12 at 3 55 17 PM

After:

Screen Shot 2022-10-12 at 3 54 51 PM

### DIFF
--- a/www/performance_optimization.php
+++ b/www/performance_optimization.php
@@ -123,9 +123,10 @@ $isMultistep = $testRunResults->countSteps() > 1;
             }
             ?>
 
+            <h3 class="hed_sub">Glossary</h3>
             <?php
-                echo '<h3 class="hed_sub">Optimization Details</h3>';
-                dumpOptimizationGlossary();
+                require_once __DIR__ . '/resources/view.php';
+                echo view('snippets.glossary', []);
             ?>
 
         </div>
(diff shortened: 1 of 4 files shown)


www/performance_optimization.php
```diff
@@ -123,9 +123,10 @@
             }
             ?>
 
+            <h3 class="hed_sub">Glossary</h3>
             <?php
-                echo '<h3 class="hed_sub">Optimization Details</h3>';
-                dumpOptimizationGlossary();
+                require_once __DIR__ . '/resources/view.php';
+                echo view('snippets.glossary', []);
```
jefflembeck, line 129: Do we want to namespace `view` so we can use `use` with it?

jefflembeck (approved):

www/performance_optimization.php
stoyan: hmm, maybe. The thing is it's like this in Laravel so I thought it'd be easier to replace.

Also `/resources/view.php` is now autoloaded in a newer PR so `view()` would be just a globally accessible function like it is in Laravel: https://laravel.com/docs/9.x/blade#introduction

PR #2466, opened 2022-10-14
Favicon

Siddhantshukla814:
Favicon updated to Circle logos


scottjehl:
This looks great. I am curious why the branch shows so many commits. Is it right ?

image

Siddhantshukla814:
I'm not sure why this happened, was there any other branch that was called favicon before? Orr did i do something wrong?

Issue #2460, opened 2022-10-12
Use the new collection period data to tell folks when their CrUX data is based on


tkadlec:
## Feature Request Summary
The CrUX API now exposes collection period data to show the aggregation window (https://twitter.com/ChromeUXReport/status/1579891834494427137) and we pull that in automatically in our JSON when we are able to retrieve CrUX data for a result.
![Screen Shot 2022-10-12 at 2 02 33 PM](https://user-images.githubusercontent.com/66536/195426498-3e89c2cd-25a7-45ef-8df4-d3763dac9f15.png)

See https://www.webpagetest.org/jsonResult.php?test=221011_AiDc1S_FCF&pretty=1 for an example.

We should use that to update our helper text on the CrUX data to say when the data was collected

![Screen Shot 2022-10-12 at 2 03 38 PM](https://user-images.githubusercontent.com/66536/195426703-cf17c2f8-f744-4c04-83df-07c4681ef338.png)

Maybe something like:

(Collected anonymously by Chrome browser via Chrome User Experience Report, between September 8, 2022 and October 11, 2022.)

tkadlec:
@sammeboy635 Wanna take a stab at it? The CrUX view is here: https://github.com/WPO-Foundation/webpagetest/blob/master/www/include/CrUX.php

sammeboy635:
Sounds good to me!

tkadlec:
Closed with #2476

PR #2476, opened 2022-10-17
Added collection period in CrUX Real User Measurement title.

sammeboy635:
#2460 
Added the collection period to CrUX Title.

I added some checks to make sure collectionPeriod is present. If it was not then it would revert back to the old way.

Here is "www.homedepot.com" test run with the WPT docker-compose.
crux


stoyan:
question: did you run `composer format` before you commit? Curious if it would see spacing like `}else{` and turn into `} else {`

sammeboy635:
@stoyan I didn't run composer format, but I just added another commit to show the differences.

stoyan (approved):

Issue #2303, opened 2022-08-25
`$experimentsPaid` vs `$experiments_paid`


stoyan:
I suspect a bug but have not confirmed. Looks like checking for `$experimentsPaid` when the global is `$experiments_paid`
https://github.com/WPO-Foundation/webpagetest/blob/5c7522d1d724bbcf0ac82f6faf64c2452210a000/www/runtest.php#L1978-L2041
and
https://github.com/WPO-Foundation/webpagetest/blob/5c7522d1d724bbcf0ac82f6faf64c2452210a000/www/runtest.php#L1749-L1763

tkadlec:
@stoyan Nice catch! Let's fix this. We've heard a couple times that people were unexpectedly getting the "Experiments are only available for WPT Pro subscribers" message. Hasn't come up a lot, but enough that it makes me think this could be what's hitting them.

PR #2477, opened 2022-10-17
fix(runtest): name vars to better validate scripts

jefflembeck:
Also switch logic (via De Morgan's Law) to make it easier to read


stoyan (approved):
addresses #2303

PR #2479, opened 2022-10-17
Blade tweaks and step 1 in optimization checklist as HTML

stoyan:
I started implementing the optimization checklist as HTML table, but a `view()` change was in order. The `gloabal` php doesn't help when the callsite is in a function, so replaced the `global $viewFactory` with a `static $viewFactory`

Also added `view.php` to autoload and fixed references.


composer.json
```diff
@@ -3,7 +3,7 @@
     "psr-4": {
       "WebPageTest\\": "www/src/"
     },
-    "files": ["www/constants.inc"]
+    "files": ["www/constants.inc", "www/resources/view.php"]
```
jefflembeck, line 6: lol this fixes the issue i mentioned before, i assume

jefflembeck (approved):

PR #2480, opened 2022-10-17
feat(acct): notify cc errors

jefflembeck:
Not the prettiest, but at least we have errors as they're passed to us. We could put together a more comprehensive and kind set of braintree errors and pass recommendations along. There's a whole list here:

https://webforcehq.zendesk.com/hc/en-us/articles/4404571172372-Braintree-Decline-Codes

### DIFF
--- a/www/templates/account/plans/plan-summary.php
+++ b/www/templates/account/plans/plan-summary.php
@@ -3,6 +3,7 @@
     <?php require_once __DIR__ . '/includes/subhed.php'; ?>
     <?php require_once __DIR__ . '/includes/billing-address-form.php'; ?>
     <!-- main form -->
+    <div id="notification-banner-container"></div>
     <form id="wpt-account-upgrade" method="post" action="/account">
         <!-- payment -->
         <div class="box card-section">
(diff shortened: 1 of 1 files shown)


www/templates/account/plans/plan-summary.php
```diff
@@ -99,4 +106,15 @@ function error(err) {
...
         });
     })();
+</script>
+<script>
+    (() => {
+        document.addEventListener('cc-upgrade-error', e => {
+            const el = document.createElement('div');
+            el.classList.add('notification-banner', 'notification-banner__error');
+            el.innerText = e.detail;
+            document.querySelector('#notification-banner-container').appendChild(el);
```
stoyan, line 116: This will continue appending new notification divs, right? Is this desirable or we want to clean up the previous error message?

stoyan (approved):

www/templates/account/plans/plan-summary.php
jefflembeck: We should probably clean it up but I want to get this out and then quickly put together something a bit better

tkadlec (approved):

PR #2486, opened 2022-10-18
Consolelog didn't have results_nav banner.

sammeboy635:
Making sure everything was working on my docker image I noticed consolelog.php didn't have the results_nav so I couldn't continue going to different results pages unless I went back. I thought it would be nice to have the same results_nav presented as the rest of them did.

Current consolelog.php
webpagetest-console

With results_nav banner consolelog.php
common-included

### DIFF
--- a/www/consolelog.php
+++ b/www/consolelog.php
@@ -1,5 +1,8 @@
 <?php
 
+include 'common.inc';
+set_time_limit(30); // Prevent the consolelogs page from running out of control.
+
 // this output buffer hack avoids a bit of circularity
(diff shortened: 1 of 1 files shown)


stoyan:
Nice!

The nav should be fixed now, addressed in a different PR I just merged https://github.com/WPO-Foundation/webpagetest/blob/master/www/consolelog.php#L2

Under what conditions the page goes out of control? Too much console.log? If so, maybe we can say "yo, too many messages, showing the first X". So we don't fail silently but still show something

PR #2487, opened 2022-10-18
Optimization checklist as an HTML table

stoyan:
* kept the image as an option to download
* reused styles from the history table
* used as opportunity to show off some Blade templating

There's definitely room for improvement both in terms of UI (e.g. old pixel-y icons) and optimization advice, but it's a start a move away from image generation

### DIFF
--- a/www/resources/views/snippets/optimizationchecklist.blade.php
+++ b/www/resources/views/snippets/optimizationchecklist.blade.php
@@ -1,1 +1,32 @@
-<!-- todo: a <table> goes here -->
+<table class="checklist">
+    <thead>
+        <tr>
+            <th>Request</th>
+            @foreach ($headers as $key => $text)
+            <th>
+                {{ $text }}
+                <br />
+                @if (isset($pageData[$key]) && $pageData[$key] >= 0 && $pageData[$key] <= 100)
+                    {{ $pageData[$key] }}%
+                @else
+                    N/A
+                @endif
+            </th>
+            @endforeach
+        </tr>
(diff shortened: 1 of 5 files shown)


stoyan:
Before: https://www.webpagetest.org/result/221007_AiDc8P_FYA/1/performance_optimization/
After:

https://user-images.githubusercontent.com/51308/196517332-be544c3c-bfc9-4d90-a702-aaca39b0430b.mov

www/resources/views/snippets/optimizationchecklist.blade.php
```diff
@@ -1 +1,32 @@
...
+                @else
+                    N/A
+                @endif
+            </th>
+            @endforeach
+        </tr>
+    </thead>
+    <tbody>
+        @foreach ($rows as $row)
+        <tr class="table-cell-{{ $row['color'] ?? ($loop->even ? 'even' : 'odd')}}">
```
jefflembeck, line 20: Where does `$loop` come from?

stoyan:
https://laravel.com/docs/9.x/blade#the-loop-variable
Blade magic

www/resources/views/snippets/optimizationchecklist.blade.php
stoyan: `$loop` is some Blade magic that helps working with loops
https://laravel.com/docs/9.x/blade#the-loop-variable

tkadlec:
Yay for HTML! And templating!

PR #2489, opened 2022-10-19
Testing out the templating with breakdownTimeline.php

sammeboy635:
@stoyan I was hoping to get some help on this if you wouldn't mind but I'm a little unsure how to PR this cleanly.
I did the bare minimum to just get breakdownTimeline.php to work with the templater and it seems to work just fine. Do you have thoughts on it?

Preview with templater
breakdown

### DIFF
--- a/www/breakdownTimeline.php
+++ b/www/breakdownTimeline.php
@@ -1,307 +1,115 @@
 <?php
-
 // Copyright 2020 Catchpoint Systems Inc.
 // Use of this source code is governed by the Polyform Shield 1.0.0 license that can be
 // found in the LICENSE.md file.
+
 include 'common.inc';
-require_once('breakdown.inc');
-require_once('contentColors.inc');
-require_once('waterfall.inc');
-require_once('page_data.inc');
 
-$page_keywords = array('Timeline Breakdown','WebPageTest','Website Speed Test','Page Speed');
-$page_description = "Chrome main-thread processing breakdown$testLabel";
+//region HEADER 
+ob_start();
+define('NOBANNER', true); // otherwise Twitch banner shows 2x
+$tab = 'Test Result';
+$subtab = 'Processing';
+include_once 'header.inc';
+$results_header = ob_get_contents();
+ob_end_clean();
+//endregion
 
-?>
-<!DOCTYPE html>
-<html lang="en-us">
-    <head>
-        <title><?php echo $page_title; ?> - WebPageTest Content Breakdown</title>
-        <script>document.documentElement.classList.add('has-js');</script>
+//region SETUP
+$processing = GetDevToolsCPUTime($testPath, $run, $cached);
 
-        <?php include('head.inc'); ?>
+if (isset($processing)) {
+    arsort($processing);
+    $mapping = array('EvaluateScript' => 'Scripting',
+                    'v8.compile' => 'Scripting',
+                    'FunctionCall' => 'Scripting',
+                    'GCEvent' => 'Scripting',
+                    'TimerFire' => 'Scripting',
+                    'EventDispatch' => 'Scripting',
+                    'TimerInstall' => 'Scripting',
+                    'TimerRemove' => 'Scripting',
+                    'XHRLoad' => 'Scripting',
+                    'XHRReadyStateChange' => 'Scripting',
+                    'MinorGC' => 'Scripting',
+                    'MajorGC' => 'Scripting',
+                    'FireAnimationFrame' => 'Scripting',
+                    'ThreadState::completeSweep' => 'Scripting',
+                    'Heap::collectGarbage' => 'Scripting',
+                    'ThreadState::performIdleLazySweep' => 'Scripting',
 
-    </head>
-    <body class="result">
-            <?php
-            $tab = 'Test Result';
-            $subtab = 'Processing';
-            include 'header.inc';
-            $processing = GetDevToolsCPUTime($testPath, $run, $cached);
-            if (isset($processing)) {
-                arsort($processing);
-                $mapping = array('EvaluateScript' => 'Scripting',
-                               'v8.compile' => 'Scripting',
-                               'FunctionCall' => 'Scripting',
-                               'GCEvent' => 'Scripting',
-                               'TimerFire' => 'Scripting',
-                               'EventDispatch' => 'Scripting',
-                               'TimerInstall' => 'Scripting',
-                               'TimerRemove' => 'Scripting',
-                               'XHRLoad' => 'Scripting',
-                               'XHRReadyStateChange' => 'Scripting',
-                               'MinorGC' => 'Scripting',
-                               'MajorGC' => 'Scripting',
-                               'FireAnimationFrame' => 'Scripting',
-                               'ThreadState::completeSweep' => 'Scripting',
-                               'Heap::collectGarbage' => 'Scripting',
-                               'ThreadState::performIdleLazySweep' => 'Scripting',
+                    'Layout' => 'Layout',
+                    'UpdateLayoutTree' => 'Layout',
+                    'RecalculateStyles' => 'Layout',
+                    'ParseAuthorStyleSheet' => 'Layout',
+                    'ScheduleStyleRecalculation' => 'Layout',
+                    'InvalidateLayout' => 'Layout',
 
-                               'Layout' => 'Layout',
-                               'UpdateLayoutTree' => 'Layout',
-                               'RecalculateStyles' => 'Layout',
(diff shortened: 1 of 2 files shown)


www/resources/views/pages/breakdownTimeline.blade.php
```diff
@@ -0,0 +1,284 @@
...
+       var tableEventsIdle = new google.visualization.Table(document.getElementById('tableEventsIdle'));
+       tableEventsIdle.draw(viewEventsIdle, {showRowNumber: false, sortColumn: 1, sortAscending: false});
+   
+       var pieEventsIdle = new google.visualization.PieChart(document.getElementById('pieEventsIdle'));
+       pieEventsIdle.draw(viewEventsIdle, {width: 450, height: 300, title: 'Processing Events', colors: eventColors});
+   }
+</script>
+
+<?php
+function rgb2html($r, $g = -1, $b = -1) // Not Used Anymore, should we keep?
```
stoyan, line 263: kill it with fire :)

www/breakdownTimeline.php
```diff
@@ -3,305 +3,28 @@
...
-          </div>
-          </div>
-          </div>
-
-
 
-        </div>
-        <?php include('footer.inc'); ?>
-        </div></div></div>
+define('NOBANNER', true); // Unclear if its needed
```
stoyan, line 7: I think it is. Do you have anything in your 'alert' setting?

https://github.com/WPO-Foundation/webpagetest/blob/d009915d76fa71d45384f5a8850ce9815fddbd68/www/header.inc#L50

if so, this is the banner that shows up on top, e.g.
Screen Shot 2022-10-21 at 12 05 39 PM

www/breakdownTimeline.php
```diff
@@ -3,305 +3,28 @@
...
-    $color .= (strlen($g) < 2 ? '0' : '') . $g;
-    $color .= (strlen($b) < 2 ? '0' : '') . $b;
-    return '#' . $color;
-}
-?>
+include 'common.inc';
+require_once('breakdown.inc');      // No affect removing
+require_once('contentColors.inc');  // No affect removing
+require_once('waterfall.inc');      // No affect removing
+require_once('page_data.inc');      // No affect removing
```
stoyan, line 13: yeah, kill 'em all

www/resources/views/pages/breakdownTimeline.blade.php
```diff
@@ -0,0 +1,284 @@
...
+@section('content')
+
+<?php 
+
+// Copyright 2020 Catchpoint Systems Inc.
+// Use of this source code is governed by the Polyform Shield 1.0.0 license that can be
+// found in the LICENSE.md file.
+
+// $page_keywords = array('Timeline Breakdown','WebPageTest','Website Speed Test','Page Speed');    // not referenced anywhere, should we keep?
+// $page_description = "Chrome main-thread processing breakdown$testLabel";                         // not referenced anywhere, should we keep?
```
stoyan, line 12: ah, maybe this SEO stuff is not yet in the new template. Maybe just keep them commented out and add "// todo: use these in the template"

stoyan (changes requested):
Sweet, I love it. I tried to address your questions inline.

www/breakdownTimeline.php
sammeboy635: I didn't have anything like that in my banner but that is probably because i'm using my docker image.

sammeboy635:
> Sweet, I love it. I tried to address your questions inline.

Hey thanks a lot for all the great feedback!

www/breakdownTimeline.php
```diff
@@ -3,305 +3,28 @@
...
+$subtab = 'Processing';
+include_once 'header.inc';
+//endregion
+
+//region template
+require_once __DIR__ . '/resources/view.php';
+echo view('pages.breakdownTimeline', [
+    //'test_results_view' => true, // Not Sure if we need this
+    'body_class' => 'result',
+    // 'results_header' => $results_header, // Null Value, Not sure if we need this
```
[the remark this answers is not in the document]
sammeboy635: image

I see what you mean, it was to fix the layout of the header. Currently without using $result_header my header looks like the image above.

Would you still recommend I pass in $result_header like you did in consolelog.php even though it is ok?
like this ?
```
ob_start();
define('NOBANNER', true); // otherwise Twitch banner shows 2x
$tab = 'Test Result';
$subtab = 'Processing';
include_once 'header.inc';
$results_header = ob_get_contents();
ob_end_clean();
```

sammeboy635:
It's ready for review :).

https://user-images.githubusercontent.com/33207323/198361041-f1ad89d1-9c21-410b-9704-2fb987eab379.mp4

www/resources/views/pages/breakdownTimeline.blade.php
```diff
@@ -0,0 +1,184 @@
...
+
+<?php 
+
+// Copyright 2020 Catchpoint Systems Inc.
+// Use of this source code is governed by the Polyform Shield 1.0.0 license that can be
+// found in the LICENSE.md file.
+
+// TODO: use these in the template
+// $page_keywords = array('Timeline Breakdown','WebPageTest','Website Speed Test','Page Speed');
+// $page_description = "Chrome main-thread processing breakdown$testLabel";
```
stoyan, line 13: You can move these  as globals to the .php file (outside the template)

They should just work. If not we'll fix in a follow up diff

www/resources/views/pages/breakdownTimeline.blade.php
```diff
@@ -0,0 +1,184 @@
...
+   
+       var events = new google.visualization.DataTable();
+       events.addColumn('string', 'Event');
+       events.addColumn('number', 'Time (ms)');
+       events.addRows(<?php echo (count($processing) - 1); ?>);
+       var eventColors = new Array();
+       <?php
+        $index = 0;
+        if (isset($groups) && is_array($groups) && count($groups)) {
+            foreach ($groups as $type => $time) {
```
stoyan, line 79: there are loops and ifs in Blade, could you try them?

https://laravel.com/docs/9.x/blade#if-statements
https://laravel.com/docs/9.x/blade#loops

the idea is to minimize `<?php` as much as we can

www/resources/views/pages/breakdownTimeline.blade.php
```diff
@@ -0,0 +1,184 @@
...
+                if ($type != 'Idle') {
+                    echo "groups.setValue($index, 0, '$type');\n";
+                    echo "groups.setValue($index, 1, $time);\n";
+                    $color = $groupColors[$type];
+                    echo "groupColors.push('$color');\n";
+                    $index++;
```
stoyan, line 85: check out the magical `$loop` var:
https://laravel.com/docs/9.x/blade#the-loop-variable

It gives you `$loop->index` :)

www/resources/views/pages/breakdownTimeline.blade.php
stoyan: And `echo` is really what we don't need when we have templates :)

www/resources/views/pages/breakdownTimeline.blade.php
```diff
@@ -0,0 +1,184 @@
...
+       tableEvents.draw(viewEvents, {showRowNumber: false, sortColumn: 1, sortAscending: false});
+   
+       var pieEvents = new google.visualization.PieChart(document.getElementById('pieEvents'));
+       pieEvents.draw(viewEvents, {width: 450, height: 300, title: 'Processing Events', colors: eventColors});
+   
+   
+       var groupsIdle = new google.visualization.DataTable();
+       groupsIdle.addColumn('string', 'Category');
+       groupsIdle.addColumn('number', 'Time (ms)');
+       groupsIdle.addRows(<?php echo count($groups); ?>);
```
stoyan, line 129: I think these count calls should work in hte template too, like

```js
groupsIdle.addRows({{ count($groups) }});
```

stoyan (changes requested):
back to your queue to see if you can remove some of the `<?php` in the template

sammeboy635:
I was going to push another PR for this but I was thinking to clean up the php we could process everything on the JS side.  This way we are not passing 200+ lines of ```groupsIdle.addRows```. Instead we just pass a json string to JS through ```var processing = {{ Illuminate\Support\Js::from($processing) }}; ```. In addition this way allows for better code reusability. Where the Timing Breakdown graph we only have to just append idle data like this
```    
    // Add Idle Data to group Table
    groupTable['data'].push(['Idle' , Groups['Idle']]);
    groupTable['colors'].push(GroupColors['Idle']);
```
Let me know if you would rather me go the larvel templating method instead.

PR #2490, opened 2022-10-20
fix(braintree): give better errors to our users

jefflembeck:
This parses braintree errors given to us by chargify and displays them for our users

### DIFF
--- a/www/assets/js/braintree-error-parser.js
+++ b/www/assets/js/braintree-error-parser.js
@@ -1,0 +1,455 @@
+(function(window) {
+  const data = {
+    "1000": {
+        "text": "Approved",
+        "implications": "The customer's card was likely approved by the network, but denied by the bank. This is usually because the provided postal code or CVV does not match the expected value"
+    },
+    "2000": {
+        "text": "Do Not Honor",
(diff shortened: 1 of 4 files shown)


www/assets/js/braintree-error-parser.js
```diff
@@ -0,0 +1,455 @@
...
+      this.code = options.code;
+      this.text = options.text;
+      this.implication = options.implication;
+    }
+  }
+
+  class BraintreeErrorParser {
+    static parse(errorText) {
+        var regExp = /\(([^)]+)\)/;
+        var errorCode = regExp.exec(errorText);
```
stoyan, line 436: nit: `const` maybe?

stoyan (approved):
sweet!

Maybe in a follow up we can make braintree-error-parser.js a module and load on demand in the event listener?

tkadlec (approved):
LGTM. 

We should definitely work on more user-friendly alternatives to the messaging for some of these, but let's get this out so at least we have more actionable information and then we can massage that after.

jefflembeck:
Yeah, i'm going to fix this up today. Will get this in and will make it nicer.

PR #2504, opened 2022-11-01
chore(plans): a little code cleanup/format

jefflembeck:
This involves a composer format but also a change where we can set available plans in config for future purposes.


tkadlec (approved):
Looks ace!!

PR #2512, opened 2022-11-03
Fixed an exception when possibly-render-blocking-reqs doesn't exist

pmeenan:
I'm assuming `possibly-render-blocking-reqs` is a custom metric on the public instance but doesn't necessarily exist elsewhere.


stoyan (approved):
nit: my personal pref is `!empty()` because `isset()` makes me think maybe the var is not declared, like it's only declared in a condition block or some such :)

PR #2522, opened 2022-11-07
Removed dynatrace references

stoyan:
We don't collect these anymore, checked with @tkadlec


pmeenan (approved):
LGTM. It's been YEARS since there were tests with dynatrace data, all long-since deleted.

PR #2524, opened 2022-11-08
Fix a subtab name

stoyan:
When the string is not consistent, the subtab menu looks off

# before

Screen Shot 2022-11-08 at 1 01 19 PM

# after

Screen Shot 2022-11-08 at 1 01 07 PM


tkadlec:
This one seems pretty straightforward. Gonna merge it in. Thanks for the cleanup, @stoyan !

Issue #1945, opened 2022-06-02
A significant portion of HTML content was generated by JavaScript after delivery on a static site


kevinfarrugia:
## Description
> A significant portion of HTML content (2kb, 7% of total HTML) was generated by JavaScript after delivery.

The above is being reported even if the HTML does not change and the only HTML changes are third-party scripts added using GTM or similar.

### Steps to reproduce
Test URL: https://www.webpagetest.org/result/220602_AiDc9J_9EZ/2/experiments/

### Expected result
Do not think it should report this as an Opportunity.

### Actual result
Getting this opportunity:
> A significant portion of HTML content (2kb, 7% of total HTML) was generated by JavaScript after delivery.

kevinfarrugia:
It would be helpful if it is possible to know what the generated HTML is or at least some examples.

rik:
Maybe this opportunity could exclude `<script>` tags from the heuristic or increase the threshold?

tkadlec:
> It would be helpful if it is possible to know what the generated HTML is or at least some examples.

This is a great suggestion! We've talked a bit about it internally too and hope to get something like this in place.

> Maybe this opportunity could exclude <script> tags from the heuristic or increase the threshold?

I think bumping the threshold might make sense here. Got suggestions on what you would think would be reasonable @rik?

rik:
I'm not sure what threshold would work best. Ignoring `<script>` would be my first try to improve the relevance of this opportunity. And after that, see if people report other false positives.

PR #2526, opened 2022-11-08
Update this wording to reflect the nuanced nature of this situation, …

scottjehl:
…which is often but not always due to javascript over-reliance

related to #1945 but not a fix


stoyan (approved):
shipit

PR #2531, opened 2022-11-09
Refactor experiments page JS to not use jQuery

### LINKED ISSUE #1939: repopulate script for experiments form can mess up freeform text encoding
spaces/+
image
Also: script should add fields to match input

scottjehl:
Additionally, fixes #1939:
- currently, when text fields repopulate from localstorage on refresh or revisiting the page, special characters were changing to url encoded ones, like spaces were pluses. Fixed.
- same scenario, multi-field experiments like find and replace weren't populating from localstorage properly. Fixed.
- same scenario, the experiments that have "add More" buttons were only populating the first input. FIxed

### DIFF
--- a/www/experiments.php
+++ b/www/experiments.php
@@ -559,19 +559,13 @@ $page_description = "Website performance test result$testLabel.";
             var pairs = priorState.split("&");
 
                 pairs.forEach(pair => {
-                    var keyval = pair.split("=");
-                    if( keyval[0]){
-                        keyval[0] = decodeURIComponent(keyval[0]);
-                    }
-                    if( keyval[1]){
-                        keyval[1] = decodeURIComponent(keyval[1]);
-                    }
+                    let keyval = pair.split("=").map(decodeURIComponent);
                     if( keyval[0] !== 'runs' ){
-                        var input = form.querySelectorAll("[type=checkbox][name='" + keyval[0] + "']");
-                        if( input.length ){
-                            input.forEach(inpt => {
-                                    if( inpt.value === keyval[1] ){
-                                        inpt.checked = true;
+                        var checks = form.querySelectorAll("[type=checkbox][name='" + keyval[0] + "']");
+                        if( checks.length ){
+                            checks.forEach(input => {
+                                    if( input.value === keyval[1] ){
+                                        input.checked = true;
                                     }
                             });
                         }
@@ -615,7 +609,7 @@ $page_description = "Website performance test result$testLabel.";
     }
 
     form.addEventListener("change", saveExperimentFormState );
-    form.addEventListener("change", saveExperimentFormState );
+    form.addEventListener("submit", saveExperimentFormState );
 
 
 
@@ -691,7 +685,6 @@ $page_description = "Website performance test result$testLabel.";
     // try and restore state at load
     refreshExperimentFormState();
     updateCount();
-    form.addEventListener("change", updateCount );
     form.addEventListener("input", updateCount );
     form.addEventListener("submit", updateCount );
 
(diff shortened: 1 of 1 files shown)


www/experiments.php
```diff
@@ -554,99 +554,170 @@ function refreshExperimentFormState(){
...
             var pairs = priorState.split("&");
 
                 pairs.forEach(pair => {
                     var keyval = pair.split("=");
+                    if( keyval[0]){
+                        keyval[0] = decodeURIComponent(keyval[0]);
+                    }
+                    if( keyval[1]){
+                        keyval[1] = decodeURIComponent(keyval[1]);
+                    }
```
stoyan, line 568: alternatively a shorter expression could be
```js
let keyval = pair.split("=").map(decodeURIComponent);
```

www/experiments.php
```diff
@@ -554,99 +554,170 @@ function refreshExperimentFormState(){
...
                     if( keyval[0] !== 'runs' ){
-                        var input = form.find("[name='" + keyval[0] + "']");
-
+                        var input = form.querySelectorAll("[type=checkbox][name='" + keyval[0] + "']");
                         if( input.length ){
-                            if( input.filter("[type=checkbox],[type=radio]").length ){
-                                input = input.filter( "[value='"+ keyval[1] +"']" ).attr("checked", true);
-                            } else if(input.filter("[type=text]").length ) {
-                                input.val(keyval[1]);
+                            input.forEach(inpt => {
```
stoyan, line 572: `inpt` reads like a typo :) Maybe make it `imp` or `i`.

Or just `input` and tweak above as
```js
let checkboxes = form.querySelectorAll("[type=checkbox]...
```

www/experiments.php
```diff
@@ -554,99 +554,170 @@ function refreshExperimentFormState(){
...
+                                    if( inpt.value === keyval[1] ){
+                                        inpt.checked = true;
+                                    }
+                            });
+                        }
+                        var input = form.querySelector("textarea[name='" + keyval[0] + "']:not([data-hydrated='true']),input[type='text'][name='" + keyval[0] + "']:not([data-hydrated='true'])");
```
stoyan, line 578: I wonder if these queries will be more readable as templates

```js
form.querySelector(`textarea[name='${keyval[0]}']:not([data-...`);

File: www/experiments.php [hunk 4]
```diff
@@ -554,99 +554,170 @@ function refreshExperimentFormState(){
...
+            formString += `${encodeURIComponent(pair[0])}=${encodeURIComponent(pair[1])}&`;
+        }
+        
+        
+        localStorage.setItem("experimentForm", formString);
     }
 
-    $("form.experiments_grades").on("change submit", saveExperimentFormState );
+    form.addEventListener("change", saveExperimentFormState );
+    form.addEventListener("change", saveExperimentFormState );
```
stoyan, line 618: duplicate?

www/experiments.php
```diff
@@ -554,99 +554,170 @@ function refreshExperimentFormState(){
...
 
-
-
-
     // try and restore state at load
     refreshExperimentFormState();
     updateCount();
-    $("form.experiments_grades").on("change input submit", updateCount );
+    form.addEventListener("change", updateCount );
+    form.addEventListener("input", updateCount );
```
stoyan, line 695: is change necessary since we're listening to `input`?

stoyan (approved):
Awesome, thanks! Few nits inline.

And don't forget to run `composer format:prettier`

www/experiments.php
scottjehl: elegant!

www/experiments.php
scottjehl: hmm yeah maybe? I left them for now... we could revisit

PR #2532, opened 2022-11-10
Add ability to sort order that experiments apply

scottjehl:
This adds sort functionality to the experiments feature so that users can control the order experiments are applied. This is helpful in cases where one experiment can destroy the changes that another made. For example, if you add a defer attribute to all scripts, and then replace the HTML of the entire page, that first change will be destroyed. The opposite order would let you replace the html then add defer.

As a first pass, this is implemented with sort buttons. We may add drag sort later if desired.

image

### DIFF
--- a/www/experiments.php
+++ b/www/experiments.php
@@ -611,7 +611,19 @@ $page_description = "Website performance test result$testLabel.";
     form.addEventListener("change", saveExperimentFormState );
     form.addEventListener("submit", saveExperimentFormState );
 
+    // append inputs to end of form on submit to impact post order
+    function sortExperimentOrder(){
+        var appliedOrder = document.querySelectorAll('.exps-active li');
+        appliedOrder.forEach(li => {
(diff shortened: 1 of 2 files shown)


www/experiments.php
```diff
@@ -611,7 +611,19 @@ function saveExperimentFormState(){
     form.addEventListener("change", saveExperimentFormState );
     form.addEventListener("submit", saveExperimentFormState );
 
+    // append inputs to end of form on submit to impact post order
+    function sortExperimentOrder(){
+        var appliedOrder = document.querySelectorAll('.exps-active li');
```
stoyan, line 616: we can switch to const/let for new code

www/experiments.php
```diff
@@ -611,7 +611,19 @@ function saveExperimentFormState(){
...
+        appliedOrder.forEach(li => {
+            var associatedInput = form.querySelector( "[name=\"" + li.getAttribute('data-input-name') + "\"][value=\"" + li.getAttribute('data-input-value') + "\"]" );
+            if(associatedInput){
+                form.append(associatedInput);
+                console.log(associatedInput);
```
stoyan, line 621: delete?

stoyan (approved):
sweet! A couple of nits inline

Issue #2519, opened 2022-11-04
Waterfall image stuck, won't scroll.


ClarkMitchell:
## Description
Waterfall image appears stuck (as if it had position: fixed), request popup and request line highlighting will scroll separately from waterfall image and no longer correspond.

### Steps to reproduce
In Chrome Canary, with a network waterfall of >250 requests, with a browser width small enough to trigger X axis overflow scrolling, click on a request item to trigger the popup. Now scroll. Popup and highlight line should scroll away, but the waterfall image stays in place.

Disabling `mix-blend-mode: multiply;` on `id=waterfall_step1`
seems to stop the bug.

The bug does not seem to occur in Safari or Firefox.

### Expected result
Waterfall Image scrolls and stays in sync with the request popup.

### Actual result
Popup and highlight line should scroll away, but the waterfall image stays stuck.

This test was used for this video demonstration: https://www.webpagetest.org/result/221104_BiDcJR_ABX/1/details/#waterfall_view_step1

https://user-images.githubusercontent.com/7844500/200014772-dd90ab8f-95b8-4255-ab80-969a66ea14c5.mp4

rik:
Is this also happening on stable Chrome? Also, I see an Update button in the top right corner, which Chrome Canary are you using?

I can't reproduce on my Chrome Canary (109.0.5400.0 arm64 macOS 13.0) nor my Chrome stable (107.0.5304.87 arm64 macOS 13.0).

scottjehl:
Following up. I think Chrome must have fixed this issue in 107 or 108...

I'll reopen if necessary but currently I'm not seeing the bug anymore. Thanks!

PR #2534, opened 2022-11-11
Removed mix-blend-mode from scrollableTable, filmstrip, waterfall...etc

### LINKED ISSUE #2533: Part of the page disappears as you scroll
## Description
When scrolling downwards on devices with a large viewport (1920 and upwards), as you scoll parts of the page get covered by a white area. This usually overlaps the requests table or the custom metrics if you have many.

### Steps to reproduce
- Browser: Google Chrome
- OS: Ubuntu 22.04

Steps:
- Go to: https://www.webpagetest.org/result/221111_AiDc1Y_64P/1/details/#waterfall_view_step1 using a device with a large viewport (in my case 1920w)
- Scroll downards, notice how the table gets covered by a white area. (see attached video)

https://user-images.githubusercontent.com/8075326/201320999-0842ec80-76fb-4817-b2a2-45c66d02e7f4.mp4

kevinfarrugia:
Fixes #2533 

Please confirm that it doesn't change anything important as I am not aware why it was added initially.

### DIFF
--- a/www/assets/css/pagestyle2.css
+++ b/www/assets/css/pagestyle2.css
@@ -6225,16 +6225,6 @@ div.overflow-container,
   background-attachment: local, local, scroll, scroll;
 }
 
-.scrollableTable table,
-.results-filmstrip-container img,
-div:not(.compare_contain_wrap) > div.waterfall-container img,
-div.overflow-container > *,
-.breakdownFrame_item > div > *,
-.experiment_opportunities-resolved ol > *,
-.experiments_details_desc > ul li {
-  mix-blend-mode: multiply;
-}
-
 .scrollableTable th,
 .scrollableTable td {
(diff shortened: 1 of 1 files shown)


rik:
This was introduced in https://github.com/WPO-Foundation/webpagetest/issues/2519

stoyan:
Thanks, Kevin! The glitchy rendering (in Chrome only) annoys me to no end. 
It does serve a purpose though - to visually show the something is scrollable. I'll defer to @scottjehl. Wondering in what other way we can accomplish the desired result without hitting the browser glitch.

kevinfarrugia:
Figured out that the mix-blend-mode: multiply doesn't play well with border-radius. Created a new PR which allows us to keep the scroll shadows. (hadn't noticed them dissappearing when pushing this PR)

PR #2535, opened 2022-11-12
HTML diff results page

stoyan:
Initial implementation of an HTML diff between HTML over the wire and `documentElement.outerHTML`. Works only is "save response bodies" is checked, otherwise we only have the rendered HTML.

The diff is performed on the client side via https://www.npmjs.com/package/diff (<8k payload)

Optionally we can format with Prettier (350k, loaded on demand only if the checkbox is checked) 

# error state

Screen Shot 2022-11-11 at 7 06 33 PM

# Normal view

Screen Shot 2022-11-11 at 7 07 06 PM

# couple of annoyances

- doctype is missing in the generated content
- there are some "inner styles set by WPT to match CSSOM " comments in the generated CSS coming from https://github.com/WPO-Foundation/webpagetest/blob/352e2aac95da976eebf1495674b6cda184da674d/www/settings/custom_metrics/rendered-html.js#L10

Not terrible when Prettier pre-formats, but otherwise pretty distracting because we lose the new lines and the diff goes wild

## without Prettier

Screen Shot 2022-11-11 at 7 21 15 PM

## with Prettier

Screen Shot 2022-11-11 at 7 20 39 PM


www/resources/views/pages/htmldiff.blade.php
```diff
@@ -0,0 +1,100 @@
...
+                    } else if (diff[i].added) {
+                        node = document.createElement('ins');
+                        node.appendChild(document.createTextNode(diff[i].value));
+                    } else {
+                        node = document.createTextNode(diff[i].value);
+                    }
+                    fragment.appendChild(node);
+                }
+                document.getElementById('diff-result').innerHTML = '';
+                document.getElementById('diff-result').appendChild(fragment);
```
stoyan, line 91: the diff rendering

scottjehl:
Looks great

scottjehl (approved):

PR #2536, opened 2022-11-14
Removed border radius from results_body

### LINKED ISSUE #2533: Part of the page disappears as you scroll

kevinfarrugia:
Fixes #2533.

From what I can tell, there is no visual impact from having the border-radius on the `results_body` so it is safe to remove. Was only visible on `min-width: 87.5em` since it was set to `0` on smaller viewports. Removing it resolves the erratic Chrome behavior with `mix-blend-mode`.

### DIFF
--- a/www/assets/css/pagestyle2.css
+++ b/www/assets/css/pagestyle2.css
@@ -582,7 +582,6 @@ span.lh-flag.poor {
 
 .results_body {
   padding: 0 3em;
-  border-radius: 1rem;
   box-shadow: 0 0 4px #ddd;
   margin: 1em -3rem !important;
   background-clip: padding-box;
--- a/www/assets/css/pagestyle2.css
+++ b/www/assets/css/pagestyle2.css
@@ -592,7 +592,6 @@ span.lh-flag.poor {
 @media (max-width: 87.5em) {
   .results_body {
     padding: 0;
-    border-radius: 0;
     box-shadow: none;
     margin: 0 0 !important;
   }


stoyan:
wow, interesting discovery, thanks!

It does look a bit different on large screens. Some screenshots for others to chime in:

## before

Screen Shot 2022-11-14 at 12 32 11 PM

## after 

Screen Shot 2022-11-14 at 12 32 19 PM

In my opinion it looks better without framing the whole thing, like on smaller screens (`(max-width: 87.5em)`). Maybe at the most a border top and bottom

Screen Shot 2022-11-14 at 12 32 24 PM

kevinfarrugia:
Do you know if the issue #2533 affects all Chrome users?

scottjehl:
Following up here. I've noticed that applying a `transform: translateZ(0)` to the overflow element (the one with `.scrollableTable` etc on it) does make the bug go away. Unsure about memory impact but it makes the pages usable

kevinfarrugia:
Seems to works well for me. :raised_hands:

stoyan:
Scott's `translateZ` is in, so closing this. Thanks again, @kevinfarrugia!

PR #2537, opened 2022-11-14
Version and move vendor JS into assets/js/vendor/

stoyan:
All versioned except `catchpoint-rum.js`, not sure what version that is.
Also deleted `dygraph-combined.js`, couldn't find any references to it.


jefflembeck (approved):
yesssss

PR #2541, opened 2022-11-16
Limit the observer effect on <style> elements

stoyan:
We currently change all `<style>` elements removing whitespace and adding a comment. This messes up the upcoming html diff feature and just in general introduces observer effect that is not always beneficial. Let's limit the use cases to only non-empty style elements. This means newly created style elements with some CSS in them. Also means we miss cases of dynamic CSS inserted into existing non-empty style elements but I'd speculate this is rare and IMO a good tradeoff compared to changing all styles.

Also add some new lines when we do change the style elements for readability.

Test page: https://www.phpied.com/files/cssom/insertRule.html
It has 4 cases:
 * regular inline style (leave it alone)
 * regular inline style that is added to later (leave it alone, missed opportunity)
 * inline style that has nothing but space (leave it alone)
 * dynamically created style with some CSS (rewrite it as before)

# original HTML (over the wire)

```html
<!DOCTYPE html>
<html>
  <head>
    <style>
      h1 {
        color: green;
      }
    </style>
    <style id="tweaked">
      h2 {
        color: blue;
      }
    </style>
    <style>
    </style>
    <title></title>
  </head>
  <body>
    <h1>Hello</h1>
    <h2>Hi</h2>
    <script>
      const el = document.createElement("style");
      document.head.appendChild(el);
      let styleSheet = el.sheet;
      styleSheet.insertRule("body { background: yellow }");

      const eel = document.getElementById("tweaked");
      styleSheet = eel.sheet;
      styleSheet.insertRule("body { padding: 66px }");
    </script>
  </body>
</html>
```

#  generated output before

```html
<html><head>
    <style>/* inner styles set by WPT to match CSSOM */h1 { color: green; }</style>
    <style id="tweaked">/* inner styles set by WPT to match CSSOM */body { padding: 66px; }h2 { color: blue; }</style>
    <style>/* inner styles set by WPT to match CSSOM */</style>
    <title></title>
  <style>/* inner styles set by WPT to match CSSOM */body { background: yellow; }</style></head>
  <body>
    <h1>Hello</h1>
    <h2>Hi</h2>
    <script>
      const el = document.createElement("style");
      document.head.appendChild(el);
      let styleSheet = el.sheet;
      styleSheet.insertRule("body { background: yellow }");

      const eel = document.getElementById("tweaked");
      styleSheet = eel.sheet;
      styleSheet.insertRule("body { padding: 66px }");
    </script>
  

</body></html>
```

# after

```html
<html><head>
    <style>
      h1 {
        color: green;
      }
    </style>
    <style id="tweaked">
      h2 {
        color: blue;
      }
    </style>
    <style>
    </style>
    <title></title>
  <style>
/* inner styles set by WPT to match CSSOM */
body { background: yellow; }
</style></head>
  <body>
    <h1>Hello</h1>
    <h2>Hi</h2>
    <script>
      const el = document.createElement("style");
      document.head.appendChild(el);
      let styleSheet = el.sheet;
      styleSheet.insertRule("body { background: yellow }");

      const eel = document.getElementById("tweaked");
      styleSheet = eel.sheet;
      styleSheet.insertRule("body { padding: 66px }");
    </script>
  

</body></html>
```


scottjehl:
Hey @stoyan!  I love the goal of this change. 

First, I do think we could avoid adding the comment, which I decided to inject merely to flag why a diff is present (say, when a style tag is suddenly populated, even though it isn't in the dom that way). I like the idea of not adding that comment.

However, I suspect that the idea of only applying the style insertion to empty `style` elements could be problematic. I'd suggest we run some experiments against sites that, in particular, use techniques like Styled Components, because this fix was directed at those specifically, and I am pretty sure there were sites that had style elements with text content and dynamic css, so this had to address those too. Some that come to mind are airbnb.com, cnn.com, twitter.com/explore

stoyan:
Thanks @scottjehl! The comment we leave is fine, it's the reformatting of CSS (missing whitespace) and some changes (colors switch between rgb() and hash syntax) that are undesirable and look like the generated HTML is way different than it is.

I tried twitter/explore and indeed the dynamic style is not empty initially, so we'll miss it with this change. Still feels awkward to mess with everyone's styles all of the time because of the people who do use dynamic styles.

Another approach that came to mind is to count the number of selectors and intervene when they are different before/after. Though I'm not sure if an API exists to count the "before" without going into some sort of parsing CSS.

stoyan:
Abandon in favor of #2562

Issue #2548, opened 2022-11-28
Swap experiments should support multiples on the runtest.php side


scottjehl:
The swap experiment is currently parsed in runtest as if it is coming from the find/replace experiment, which uses swap, but that experiment actually is an edge case that uses a few fields to build a swap experiment value. Instead, let's rename that find/replace experiment to something unique like "findreplace" so that runtest can handle it specifically without messing with other swap experiments that are in the form (meta viewport is one).

PR #2550, opened 2022-11-28
Add Opportunity and Experiment for images that are too large for display

scottjehl:
Note: not ready until we decide on an image API service, which will be set via settings. 
Also: This PR builds upon #2548

This new opportunity detects images with natural widths that are more than twice their display width in the page layout. While it does not go further to determine whether the image is sized well or not across other viewport sizes, it does give an idea of how the images are targeted to the particular environment that was tested.

In addition to detecting images that are more than twice as large as displayed, this PR offers a WebPageTest Pro experiment that will resize these images to the width they will be displayed. Such an experiment can be useful in determining the benefits of implementing responsive image markup and serving multiple image sizes.

In the PR, the particular image service URL set via settings. We'll need to investigate the ideal API and whether or not we use an internal one or a third party service.

### DIFF
--- a/www/experiments/imgs-wasteful.inc
+++ b/www/experiments/imgs-wasteful.inc
@@ -1,0 +1,57 @@
+<?php
+
+$category = "Quick";
+$imgsTooLarge = $testStepResult->getMetric('imgs-too-large');
+
+function fullURL($url)
+{
+    global $requests;
+    if (preg_match('/^(https?\:\/\/)/', $url)) {
+        return $url;
+    } else {
+        return $requests[0]['url'] . $url;
+    }
+    return $url;
+}
+
+if (count($imgsTooLarge)) {
+    $tooLargeList = array();
+    $experimentSwapList = array();
+    foreach ($imgsTooLarge as $img) {
+        $srcToUse = "";
+
+        if ($img["currentSrc"]) {
+            $srcToUse .= $img["currentSrc"];
+        } elseif ($img["src"]) {
+            $srcToUse .= $img["src"];
+        }
+        $thisListItem = $srcToUse . " (Display: " . $img["displayWidth"] . "px wide, File: " . $img["naturalWidth"] . "px wide)";
+        $thisSwapItem = encodeURIComponent($srcToUse) . "|" . encodeURIComponent("{IMAGE_SERVICE_URL_HERE}/w_" . $img["displayWidth"] . "/" . fullURL($srcToUse));
+        array_push($tooLargeList, $thisListItem);
+        array_push($experimentSwapList, $thisSwapItem);
+    }
+
+    $assessment[$category]["opportunities"][] = array(
+        "title" =>  "Images are larger than necessary for display",
+        "desc" => "When images are much larger than they appear in the page, they increase download size without noticeably improving quality.",
+        "examples" => array_unique($tooLargeList),
+        "experiments" =>  array(
+            (object) [
+                "id" => '054',
+                'title' => 'See potential impact of using responsive images',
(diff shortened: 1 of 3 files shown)


www/settings/custom_metrics/imgs-too-large.js
```diff
@@ -0,0 +1,34 @@
+var imgsTooLarge = [];
```
stoyan, line 1: I'd say switch to `let`/`const` for new code

www/settings/custom_metrics/imgs-too-large.js
```diff
@@ -0,0 +1,34 @@
...
+var allImgs = document.querySelectorAll("img");
+var widthTolerance = 0.5;
```
stoyan, line 3: is it worth checking height too?

www/settings/custom_metrics/imgs-too-large.js
```diff
@@ -0,0 +1,34 @@
...
+  //display none isn't enough (only catches if that image is display non, not parent)
+  //so we check offsetParent, but also position fixed (in everyone but ff, position fix makes offsetParent null)
+  let isVisible =
+    window.getComputedStyle(allImgs[i]).display != "none" &&
+    (allImgs[i].offsetParent != null ||
+      window.getComputedStyle(allImgs[i]).position == "fixed");
+  // if display width is less than 50% of the full image natural size
+  if (
+    isVisible &&
+    boundingClientRect.width < allImgs[i].naturalWidth * widthTolerance
```
stoyan, line 16: This is going to catch svgs too, right? Where `naturalWidth` is way higher but that's probably ok.

Probably not worth special-casing SVG since the goal is to experiment

www/settings/custom_metrics/imgs-too-large.js
```diff
@@ -0,0 +1,34 @@
+var imgsTooLarge = [];
+var allImgs = document.querySelectorAll("img");
+var widthTolerance = 0.5;
+for (var i = 0; i < allImgs.length; i++) {
+  var boundingClientRect = allImgs[i].getBoundingClientRect();
+  //getBoundingClientRect returns 0 if the image is hidden (display none or parent display none)
+  //display none isn't enough (only catches if that image is display non, not parent)
+  //so we check offsetParent, but also position fixed (in everyone but ff, position fix makes offsetParent null)
+  let isVisible =
+    window.getComputedStyle(allImgs[i]).display != "none" &&
```
stoyan, line 10: prefer `!==` and `===`

www/settings/custom_metrics/imgs-too-large.js
```diff
@@ -0,0 +1,34 @@
+var imgsTooLarge = [];
+var allImgs = document.querySelectorAll("img");
```
stoyan, line 2: Might be cool to inspect background images too, in a follow up diff maybe

www/experiments/imgs-wasteful.inc
```diff
@@ -0,0 +1,73 @@
+<?php
+
+$category = "Quick";
+$imgsTooLarge = $testStepResult->getMetric('imgs-too-large');
+$imgResizeURL = Util::getSetting('imgResizeURL', null);
+
+function fullURL($url)
+{
+    global $requests;
+    if (preg_match('/^(https?\:\/\/)/', $url)) {
```
stoyan, line 10: probs better to use https://php.net/parse_url, something like
```php
if (parse_url($url, PHP_URL_SCHEME)) {

stoyan (approved):
Looks great, few nits inline.

www/experiments/imgs-wasteful.inc
scottjehl: good tip thanks

www/settings/custom_metrics/imgs-too-large.js
scottjehl: yeah, with bg size I guess? harder to parse but I'm interested

PR #2562, opened 2022-11-30
New metric to minimize the observer effect on CSS

stoyan:
* like rendered-html but with doctype and less aggressive <style> rewrite
* count curly braces as an idea if a CSS was modified
* works on a clone of the doc so we don't mangle <style> twice
* run before rendered-html which does mangle <styles>
* delete the doctype metric since we have no use for it after this diff
* only collect the metric if the response bodies are gathered, otherwise we cannot html diff anyway


scottjehl:
Looks great

scottjehl:
@stoyan I wonder if we should add the doctype over at the other rendered-html metric now that we have it?

PR #2563, opened 2022-12-01
Don't recommend priority hints if they're already there

### LINKED ISSUE #2544: priority hints are recommended on LCP images that have them already
Example
https://www.webpagetest.org/result/221103_BiDc1C_FWP/2/experiments/#Quick

scottjehl:
Fixes #2544

### DIFF
--- a/www/experiments/lcp.inc
+++ b/www/experiments/lcp.inc
@@ -23,9 +23,10 @@ if (isset($lcp)) {
     if ($lcp['time'] > 2500) {
         $isBG = false;
         $isVideo = false;
-
+        $hasHighFetchPriority = false;
         if (isset($lcp['element']['outerHTML'])) {
             $lcpHTML = '<code class="language-html">' . htmlentities($lcp['element']['outerHTML']) . '</code>';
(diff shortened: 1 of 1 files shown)


jefflembeck (approved):

www/experiments/lcp.inc
```diff
@@ -74,8 +75,9 @@ if (isset($lcp)) {
                 "expval" => array($lcpSource . "|as_image"),
                 "explabel" => array($lcpSource)
             ];
+            
             // priority Hints only help for foreground images
-            if (!$isBG && !$isVideo && !(isset($lcp['element']['fetchPriority']) && $lcp['element']['fetchPriority'] == 'high')) {
+            if (!$isBG && !$isVideo && !$hasHighFetchPriority) {
```
jefflembeck, line 80: This is getting close to breaking out into its own variable or changing into something like 

```php
$needs_priority_hints = !($isBG || $isVideo || $hasHighFetchPriority)
```

scottjehl: hmm yeah nice idea. maybe `$suggest_priority_hints = ` ?

PR #2614, opened 2023-01-03
fix(signup): button size too small on safari

jefflembeck:
No font-size set means we're stuck with browser vendor defaults and safari says "small"

Before:

before-signup-button

After:

after-signup-button


stoyan (approved):
shipshipship

tkadlec (approved):
yes please!!

PR #2615, opened 2023-01-04
Don't count LH as a test run if standard WPT test is run

tkadlec:
We're no longer going to be counting Lighthouse runs as a separate run when selected alongside a standard WebPageTest result.

This change makes sure that if a standard test is run, running LH has no change on the test run count. However if only LH is run, it should count against users test run limits.


jefflembeck (approved):
🐑

PR #2619, opened 2023-01-05
Fix Login on LH results

stoyan:
Looks like we were including `common.inc` twice which probably resets some checks.

As a result the page loads in signed out state


stoyan:
btw, I chose to `_once` rather than delete the include because lighthouse-new.php works as a standalone page too

jefflembeck (approved):

PR #2628, opened 2023-01-06
fix(api keys): disable button on submit

jefflembeck:
Sometimes the request takes a second to process and that leads to a user maybe pushing it twice or maybe it takes a second and they double click or maybe the browser acts a fool and is like "yes absolutely submit this form twice" idk, things happen out there ¯\_(ツ)_/¯

Anyway, this prevents that mess.


stoyan (approved):
I wonder if it's even necessary to `preventDefault()` and then `submit()`. But this looks good either way.

Issue #2620, opened 2023-01-05
Remove "capture video" checkbox


stoyan:
## Feature Request Summary
Looks like it's being ignored and video is captured anyway. Also we need video for some metrics (Start Render, Speed Index) and features (filmstrip?)

### What's the motivation or use case for the change/feature?
Cleaner UI and codebase

Tagging @pmeenan for thoughts

pmeenan:
Not sure if it's linked to that checkbox specifically (and it won't really hurt to remove it) but the agents support capturing video for calculating the metrics but not uploading it so you can still get the visual metrics without the storage requirements of having to keep filmstrips.

PR #2629, opened 2023-01-09
Remove "Capture Video" option

stoyan:
* See #2620
* The videos are small and we really like the filmstrips
* Only removed the UI checkbox, left the batch and API options for now


tkadlec (approved):
🐑

PR #2631, opened 2023-01-09
Show subitems in LH details

stoyan:
Moved the details for its own Blade component to abstract the `type` switching.

One slightly irky thing is that we have empty cells now whenever items and subitems have different tds in the tr. LH gets around by not showing any vertical lines in tables.

## before the change

Screenshot 2023-01-09 at 2 49 46 PM

## after

Screenshot 2023-01-09 at 2 49 22 PM

## LH

Screenshot 2023-01-09 at 2 49 59 PM


stoyan:
updated screenshot 
Screenshot 2023-01-09 at 4 08 52 PM

PR #2635, opened 2023-01-10
Remove Bulk file uploads in favor of File Reader

stoyan:
* consistent with the other textareas where we allow them to be populated from a text file
* client side, no uploads

### DIFF
--- a/www/assets/js/test.js
+++ b/www/assets/js/test.js
@@ -16,8 +16,7 @@ function ValidateInput(form, remainingRuns) {
   if (
     (form.url.value == "" || form.url.value == "Enter a Website URL") &&
     form.script.value == "" &&
-    (form["bulkurls"] == undefined || form.bulkurls.value == "") &&
-    (form["bulkfile"] == undefined || form.bulkfile.value == "")
+    (form["bulkurls"] == undefined || form.bulkurls.value == "")
   ) {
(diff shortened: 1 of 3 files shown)


stoyan:
@pmeenan, wdyt? It should not affect API call, correct?

stoyan:
## before
Screenshot 2023-01-10 at 4 59 50 PM

## after
Screenshot 2023-01-10 at 4 59 14 PM

pmeenan:
Should be ok.  Technically, it affects the api since runtest.php no longer handles bulkfile but I'd be shocked if anyone used the API for running bulk testing that way (and even more shocked if they used bulkfile).  If you're using the API already, odds are you are sending each URL as a separate API call.

pmeenan:
If you want to be particularly safe, remove the UI side of things but leave the handling in runtest.php alone.

stoyan:
>  If you're using the API already, odds are you are sending each URL as a separate API call.

yeah, or use the `bulkurl`

Also the API docs do not contain the word "bulk" :)

pmeenan:
The one I was less sure about was the bulk tester google doc that Andy put together but it looks like it also uses the API for individual URLs: https://github.com/WebPageTest/WebPageTest-Bulk-Tester/blob/master/webpagetest.gs

I'd go ahead and remove it.

PR #2641, opened 2023-01-12
Refactor opportunities & experiments includes

stoyan:
* Refactoring due to a recent global var mishap
* Add `AssesmentRegistry` singleton
* Categories are now constants
* Each opp calls `register()` to add an opportunity
* Each opp is now defined in an anonymous function (where needed) as to not bleed temporary vars as globals
* Using e.g. `global $testStepResult` makes it obvious where we have further space to improve, e.g. memoize the function that returns test step result and call it when needed, as opposed to rely on a global

Usage:

```php
// get the single registry object
$reg = AssesmentRegistry::getInstance();

// register a new opportunity
$reg->register(
  AssesmentRegistry::Quick,  // category
  ['title' => 'etc...'], // actual opp
);

// use 'em all in common.inc
$assesments = $reg->getAll();
```


scottjehl:
I'm pouring over this incredibly thorough and complete PR and I guess my only feedback so far is a minor variable edit change: `assesment ` -> `assessment`? 

This looks really great, @stoyan

scottjehl (approved):

jefflembeck (approved):
YES

Issue #2471, opened 2022-10-14
Add Experiment: Edit response body


scottjehl:
This experiment allows you to edit the response of any same-domain file requested in this test. The text in this field will replace the text of the response before it is sent to the browser.

As an initial release, allowing editing the initial request response body is likely the most important one (the HTML).
Ideally, we'd allow one or many of these fields and offer all same domain requests in the select menu.

## Proposed Design

<img width="1381" alt="image" src="https://user-images.githubusercontent.com/214783/195931107-dd098ee5-3011-4472-8e95-9757fa239b22.png">

scottjehl:
Merged!

PR #2643, opened 2023-01-12
Edit Initial Response HTML experiment. #2471 

scottjehl:
A proof of concept for review.

This is the experiments page UI to pair with a new Edit HTML Response body experiment. It allows WebPageTest Pro users to freely edit the response HTML of the site being tested, and the experiment runner will deliver that edited HTML directly to the browser as if it was server-generated.

The big advantage here is control over complicated changes to HTML, which is going to be preferable in many cases to running a find and replace experiment.

Here's how it looks in the Opportunities page on any test that has a response body for the initial HTML:
image

Currently at least, on focus, the textarea fills much of the window. We'll need to think about this toggle from an accessibility and usability standpoint, but it's nice for big text edits I think.
image

This experiment is a bit different than others in that the experiment post data can be quite large, so the experiment runner does not expect to find that posted HTML in the cookie we usually send for experiments, but instead in the test's metadata. There's an endpoint set up in the files of this PR that gets that HTML from the metadata. The experiment proxy fetches that HTML using the test ID that is submitted as a cookie and uses that response as the HTML response.

Here's a result of an experiment that used an completely different HTML source with just an H1 in it:
image

Still to-do:
- review experiment post/retrieve workflow to see if it's a good approach for experiments that rely on heavier post data
- limit the experiment proxy to the specific request url instead of any HTML-type request as it currently replaces.
- consider accessibility of textarea expand
- Note: This experiment relies upon a pull request in the experiments repo, so that'll need to merge first.


www/experiments/getExperimentHTML.php
```diff
@@ -0,0 +1,19 @@
+<?php
+require_once '../common.inc';
+use WebPageTest\Util;
+require_once '../include/TestInfo.php';
+$testInfo = json_decode(gz_file_get_contents("../$testPath/testinfo.json.gz"));
+
+if($testInfo && $testInfo->metadata){
+    $recipes = json_decode($testInfo->metadata)->experiment->recipes;
+    $html = "";
```
stoyan, line 9: move outside the `if` since the last `echo` is also outside it

www/experiments/custom.inc
```diff
@@ -68,6 +72,26 @@ $assessment[$category]["opportunities"][] = array(
...
+        "title" =>  "Edit Response HTML",
+        "desc" =>  "This experiment allows you to freely edit the text of the initial HTML response",
+        "examples" =>  array(),
+        "experiments" =>  array(
+            (object) [
+                'id' => '054',
+                'title' => 'Edit Response HTML',
+                "desc" => '<p>This experiment will replace the initial HTML with the submitted value as if it was served that way initially.</p>',
+                "expvar" => 'editresponsehtml',
+                'textinputvalue' => htmlentities(file_get_contents("https://www.webpagetest.org" . $requests[0]['body_url'])),
```
stoyan, line 86: hmm, isn't there a better way to fetch the body? We can address this todo maybe: 
https://github.com/WPO-Foundation/webpagetest/blob/2fabc42e7a87af394f363098ac01ad89bd9f4fd7/www/htmldiff.php#L39

www/runtest.php
```diff
@@ -1109,7 +1113,7 @@ function buildSelfHost($hosts)
 
 
                         //replace last step with last step plus recipes
-                        $test['script'] = str_replace($scriptNavigate, "setCookie\t" . $originToUse . "\twpt-experiments=" . urlencode($recipeScript) . "\r\n" . $scriptNavigate, $test['script']);
+                        $test['script'] = str_replace($scriptNavigate, "setCookie\t" . $originToUse . "\twpt-experiments=" . urlencode($recipeScript) . "\r\n" . "setCookie\t" . $originToUse . "\twpt-testid=" . urlencode($id) . "\r\n" . $scriptNavigate, $test['script']);
```
stoyan, line 1116: are duplicating `"setCookie\t" . $originToUse`?

stoyan: oh, I got it, nevermind

www/experiments/custom.inc
scottjehl: Would love to see that moved to a utility sure.

PR #2648, opened 2023-01-16
Breakpoint fix for footer ul

Siddhantshukla814:
Shifted breakpoint from 75em -> 85em, this skips the footer layout problem.


scottjehl (approved):
Nice!

PR #2649, opened 2023-01-17
Signup FAQ line width adjustment

Siddhantshukla814:
Max width is set to 55em, provides a slightly better reading experience.

image


scottjehl (approved):
Great! Thanks

PR #2653, opened 2023-01-18
Load the details page with display_errors 1

stoyan:
Added `idx()` helper and fixed whatever notices appeared while loading http://www.,webpagetest.org/result/230105_AiDcJS_E7F/1/details/


www/helpers.php
```diff
@@ -0,0 +1,7 @@
+<?php
+function idx($array, $key, $default = null){
```
jefflembeck, line 2: I _think_ this function is equivalent to:

```php
return $array[$key] ?? $default;
```

stoyan: yeah, you're right, this helper is a little much and not really necessary

stoyan:
Reverted idx() and composer/vendor changes. Still no errors or notices on webpagetest.org/result/230105_AiDcJS_E7F/1/details/

jefflembeck (approved):

PR #2670, opened 2023-01-23
Timings table update

stoyan:
* Remove `domInteractive` as it's virtually equivalent to `domContentLoadedEventStart` that's right next to it. Less clutter.
* Rename `domContentLoaded` to `domContentLoadedEvent` to be consistent with `loadEvent` next to it. So the logic for both is that`*Event` stands for W3C timing's `*EventEnd - *EventStart`
* add a `title` for context

## before

Screenshot 2023-01-23 at 3 22 11 PM

## after

Screenshot 2023-01-23 at 3 21 56 PM


stoyan:
This and a lot more in #2678

PR #2678, opened 2023-01-25
Revamp Timings table

stoyan:
* Add `CustomMetricFiles`, reused with runtest.php to find the "hardcoded" metric gathering (those in .js files)
 * removed custom metrics that are hardcoded (such as Colordepth) leaving only the ones set up by the user in this particular test
 * metrics go with their step #
 * group metrics by source (element, user...)
 * headers are links to docs (one w3c, two mdn, one wpt doc)
 * data massaging is in php classes and presentation in blade

## before

Screenshot 2023-01-25 at 4 34 27 PM

## after

Screenshot 2023-01-25 at 4 34 12 PM


www/details.php
```diff
@@ -113,7 +112,8 @@ function createForm($formName, $btnText, $id, $owner, $secret)
 
                         echo "<a href=\"$link\"" . ($run === $i && !$cached ? ' aria-current="page"' : '') . ">Run $i</a>";
                         if ($linkCACHED) {
-                            echo " <a href=\"$linkCACHED\"" . ($run === $i && $cached ? ' aria-current="page"' : '') . ">(Repeat View)</a>";;
+                            echo " <a href=\"$linkCACHED\"" . ($run === $i && $cached ? ' aria-current="page"' : '') . ">(Repeat View)</a>";
+                            ;
```
jefflembeck, line 116: Seems unnecessary

```suggestion
```

www/resources/views/partials/timings.blade.php
```diff
@@ -0,0 +1,55 @@
...
+                @foreach ($data as $category => $items)
+                @foreach ($items as $itemcategory => $_)
+                <th>
+                    @if ($category === 'navigation')
+                    <span title="Time it took for all event handlers to run">{{ $itemcategory }}</span>
+                    @else
+                    {{ $itemcategory }}
+                    @endif
+                </th>
+                @endforeach
```
jefflembeck, line 38: Little thing but does our formatting not pick up on this indent? (I missed while reading because it feels like the indent should nest here)

stoyan: yeah, that irks me too, the VS code formatter doesn't indent these and and puts them on the same column. I end up doing it by hand only to have them undone if I'm not careful, like:
```html
@foreach
@if
<div/>
@endif
@endforeach
```
The php formatter doesn't know about `@`s so ignores them

stoyan:
left-aligned table headers
Screenshot 2023-03-09 at 11 37 34 AM

www/resources/views/partials/timings.blade.php
stoyan: Reformatted the offenders https://github.com/WPO-Foundation/webpagetest/pull/2678/commits/951f8926ec62a27bd2784ac0f2b4a8a34f3e3f68

PR #2679, opened 2023-01-26
fix(account): change how isPaid works

jefflembeck:
Users who are actually cancelled are free users now. Let's treat them as such

Also, there is some formatting in this


stoyan (commented):
lgtm

PR #2697, opened 2023-02-07
feat(psalm): typecheck the things!

jefflembeck:
Psalm helps us by finding spots in our code where we have typing issues (using things we shouldn't, or are missing the tags that we need to use our tooling correctly)


stoyan (commented):
this is awesome!

Next step is having fewer `array|false|string` ambiguities :)

stoyan (approved):

PR #2698, opened 2023-02-07
feat(contact): add vat_number to update

jefflembeck:
This allows a user to enter their VAT number, which is a requirement on
an invoice in many countries


www/src/CPGraphQlTypes/ChargifySubscriptionInputType.php
```diff
@@ -12,22 +12,34 @@ class ChargifySubscriptionInputType
...
+        string $plan_handle,
+        string $payment_token,
+        ChargifyAddressInput $address,
+        ?string $vat_number = null
+    ) {
         $this->plan_handle = $plan_handle;
         $this->payment_token = $payment_token;
         $this->shipping_address = $address;
         $this->billing_address = $address;
+        $this->vat_number ??= $vat_number;
```
stoyan, line 27: `??=` wow, look at that, #til :D

jefflembeck: Ya know, I love `??=` but you mentioning it here reminds me that I don't think there's any situation where I would need it here. It's basically "is this set? if not, then go with this other thing" but this is in the constructor.

jefflembeck, line 27:
```suggestion
        $this->vat_number = $vat_number;
```

PR #2702, opened 2023-02-08
feat(api): add private test

jefflembeck:
This allows an api user to create a test that is truly private, meaning it can only be accessed by the owner


stoyan (approved):
lgtm

PR #2717, opened 2023-02-24
fix(account): better error for invoice

jefflembeck:
We should have a good message for users on their invoice page if they need an invoice from before when our billing system auto generated


stoyan (approved):
looks good

PR #2724, opened 2023-02-28
fix(results): low priority tests are not private

jefflembeck:
This is a hack. This is definitely a hack. We had a situation where anon users were creating tests that were set to private by default. Unfortunately, anon users cannot access private tests. That's just not how that works.


tkadlec (approved):

stoyan (approved):
shipit

PR #2732, opened 2023-03-06
fix(services): shorten timeout for any calls out to CP

jefflembeck:
right now we have a 10 minute timeout. when scheduler goes down, everything goes kaboom.


tkadlec (approved):
🚀

PR #2754, opened 2023-03-08
these are used in common. included files will set them

scottjehl:


scottjehl:
Verbally okay'd by @jefflembeck :)

PR #2785, opened 2023-03-15
feat(auth): add user ids to tests

jefflembeck:
Break contact id out into its own thing, use user ids for gating access to tests for the time being, as well.

This is a hold over until we get a solid authz service in place


stoyan (approved):
looks good!

PR #2792, opened 2023-03-17
Add Server Timing to home page

tkadlec:
Our TTFB sometimes has some issues, so this adds a class for setting Timers that we can then output in the Server-Timing Header. At the moment, this is only adding them to the homepage.


www/src/Util/Timers.php
```diff
@@ -0,0 +1,47 @@
+<?php
```
jefflembeck, line 1: Feels like this format for the header is supposed to be 

`{$description};dur={$duration}`

jefflembeck (approved):
This is real neat. Excited to use this on all sorts of stuff.

jefflembeck:
Might be worth locking down some of these metrics, though (logged in users? admins?)

mjkozicki (approved):

PR #2796, opened 2023-03-22
fix metaInfo setter, which brings back experiments nav links

scottjehl:


stoyan (approved):
looks good, sorry for breaking it

PR #2808, opened 2023-03-30
New blog posts for the homepage

stoyan:
Screenshot 2023-03-30 at 1 36 43 PM


scottjehl:
good idea. These need images that fit the same dimensions so they line up vertically

stoyan:
Here's a version where we use all images from the blog posts and they align, it's just the first one feels odd as it's a screenshot of the same page
Screenshot 2023-04-05 at 11 35 18 AM

PR #2809, opened 2023-03-31
Adding Moto G Power

tkadlec:
Lighthouse recently changed their default device to the Moto G Power (https://github.com/GoogleChrome/lighthouse/pull/14674). This means we have a situation where our default device (the G4) has a much smaller viewport, causing some sizable differences in what folks may see in our LH runs and WPT runs.

This adds the Moto G Power. Using a 7.5x CPU throttling factor seems about right.

A referenced Speedometer 2.0 benchmark on a real device came in at 14.1 +/- 0.21. A few quick tests at 7.5 returned: 14.5, 13.9, 14.1

### DIFF
--- a/docker/local/server/config/mobile_devices.ini
+++ b/docker/local/server/config/mobile_devices.ini
@@ -8,7 +8,7 @@ group=Android
 width=412
 height=767
 dpr=1.75
-throttle_cpu=7
+throttle_cpu=7.5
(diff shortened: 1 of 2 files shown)


tkadlec:
At _some_ point, we should look at making this the default. But I think we want to be careful about communicating it since anyone running default tests without specifying the mobile device _will_ see different metrics than they're used to.

stoyan (approved):

PR #2823, opened 2023-04-11
Remove viewsqlite.php and any SQLite leftovers

stoyan:
Quick fixes:
 * isset() was causing a 500
 * looping over null $results was causing an error too

Long term question:
 * Delete this completely?


stoyan:
Turns out this is a leftover from an old system of labels.

There was one place that was still working: in the compare page there's an option to edit a label. 

Screenshot 2023-04-12 at 3 12 26 PM

This label is written to /dat/labels.db 

There are few problems with this:
 * privacy: I changed a label of a test I didn't run. It persisted in incognito, so it's changed for everyone
 * security: admittedly a low level but an `alert()` in the title executed the script while modifying the label. After a refresh it was escaped properly
 * general confusion: the label is changed only on the compare page, the test otherwise shows the original label

Overall I love the idea of changing a label post-factum, but maybe a better implementation is in order

pmeenan (approved):
LGTM, thanks for the cleanup

PR #2824, opened 2023-04-13
Refactor testlog.php

stoyan:
* move admin interface to Blade
 * better separation of results retrieval (log grepping moved to `getHistoryLog()`) and presentation
 * untangle CSV conditions
 * minor CSS tweaks

# After
Screenshot 2023-04-13 at 5 36 26 PM

# Before
Screenshot 2023-04-13 at 5 28 32 PM

Also tested CSV results to see they're identical

### DIFF
--- a/www/testlog.php
+++ b/www/testlog.php
@@ -55,8 +55,8 @@ if (!$csv && ($is_logged_in || (!isset($user) && !isset($_COOKIE['google_email']
 
     ];
 
-    //echo view('pages.testhistory', $vars);
-    //exit();
+    echo view('pages.testhistory', $vars);
+    exit();
 }
 
 //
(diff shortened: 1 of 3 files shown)


www/testlog.php
```diff
@@ -7,55 +7,75 @@
 
 use WebPageTest\Util;
 
-$host = Util::getSetting('host');
-$protocol = $request_context->getUrlProtocol();
+// why are you even here
+if ($userIsBot || Util::getSetting('disableTestlog')) {
+    header('HTTP/1.0 403 Forbidden');
```
jefflembeck, line 12: can we instead throw a Forbidden exception here?

These get handled here: https://github.com/WPO-Foundation/webpagetest/blob/master/www/common.inc#L41


I like giving 404s back to bots and people who don't know much better. We can always log this stuff in our exception handler

stoyan: nice, #til

jefflembeck (approved):

PR #2831, opened 2023-04-18
Add link to addresses.php in admin_header.inc

stoyan:
Looks like `addresses.php` includes `admin_header.inc` so it should be part of the whole admin interface


www/admin_header.inc
```diff
@@ -41,6 +41,7 @@ function header_link($url, $title)
                     <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
                         <ul class="nav navbar-nav">
                             <?php header_link('/getLocations.php?f=html', 'Locations');?>
+                            <?php header_link('/adresses.php', 'Location IPs');?>
```
jefflembeck, line 44:
```suggestion
                            <?php header_link('/addresses.php', 'Location IPs');?>
```

jefflembeck (commented):
Just one thing there

PR #2838, opened 2023-04-19
Fix forcePrivate option so it really prevents the all option 

fbuecklers:
Hi,

I see that you have recently refactored the test log and reimplemented the forcePrivate option. However, the option seems to be still broken since it does not prevent any user from appending &all=on to the teslog.php 

This will still show the all tests regardless if the forcePrivate option was configured. This PR will fix that.


jefflembeck (approved):

jefflembeck:
Thanks for the catch!

fbuecklers:
Cool, thank you for the fast reply and merge!

PR #2853, opened 2023-04-24
Blade-ify blocked.php. Also fixes white-on-white header.

stoyan:


stoyan:
The verbiage about youtube being blocked doesn't seem to be true anymore
https://www.webpagetest.org/result/230424_AiDcVF_RRH/

PR #2863, opened 2023-05-01
fix(account): logout bad clients 

jefflembeck:
When a user's account is canceled, we moved them to the free client and
suspend their paid one. This will cause issues for them when they're
using the site and will continue to cause problems with them in the
future unless we log them out so they log back in with a better
contact/client for their actual account status.

Note: this cannot go out to prod until post release and account management that makes this possible


stoyan (approved):
lgtm

Issue #2826, opened 2023-04-15
ffmpeg in Docker local


daneah:
## Description
When running the local server/agent with Docker Compose on macOS with an M1 chip, the test for ffmpeg fails

### Steps to reproduce

1. Be on macOS with an M1 chip
2. Clone this repo
3. Run `sudo docker-compose up`
4. Visit `http://localhost/install`
5. Observe failing ffmpeg test
6. Observe logs:

<details><summary>Logs</summary>

```
webpagetest-php-1    | Error loading shared library librav1e.so.0: No such file or directory (needed by /opt/ffmpeg/bin/ffmpeg)
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_twopass_in: symbol not found
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_data_unref: symbol not found
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_receive_packet: symbol not found
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_config_default: symbol not found
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_config_parse_int: symbol not found
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_config_unref: symbol not found
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_context_unref: symbol not found
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_context_new: symbol not found
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_send_frame: symbol not found
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_frame_new: symbol not found
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_status_to_str: symbol not found
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_config_set_pixel_format: symbol not found
webpagetest-php-1    | Error relocating /opt/ffmpeg/bin/ffmpeg: rav1e_frame_fill_plane: symbol not found
[...]

PR #2865, opened 2023-05-02
fix docker local ffmpeg fdk-aac-dev and librav1e not found

dbeneker:
Couldn't get the local docker variant to work because ffmpeg can not find `fdk-aac-dev` and `librav1e`.
This resolves the issues and should also resolve https://github.com/WPO-Foundation/webpagetest/issues/2826


jefflembeck:
Thanks!

PR #2873, opened 2023-05-03
Fix cached run check

stoyan:
`$this->runResults->isCachedRun()` is always `false` as `runResult` is always the first result

`$cachedRun = $stepResult->isCachedRun();` is the correct one done is `_createRow()` so use that to propagate through headers and footers

Tested in FV-only test https://www.webpagetest.org/result/230310_AiDcXP_BJ9/
and in RV too https://www.webpagetest.org/result/230502_AiDcCW_B19/

### DIFF
--- a/www/include/RunResultHtmlTable.php
+++ b/www/include/RunResultHtmlTable.php
@@ -153,7 +153,7 @@ class RunResultHtmlTable
         return $out;
     }
 
-    private function _createHead()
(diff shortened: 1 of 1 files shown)


stoyan:
(cherry picked here: https://github.com/WPO-Foundation/webpagetest/commit/62894c3653f6b92771341d61ec6115e3094816e0)

PR #2890, opened 2023-05-04
Simplify `InsertCruxHTML()`

stoyan:
Not sure why Web Vitals page doesn't have all the CrUX metrics. If any page should have them all, that's the one :)

Also some cleanup as a side effect

### DIFF
--- a/www/vitals.php
+++ b/www/vitals.php
@@ -160,7 +160,7 @@ function InsertWebVitalsHTML($stepResult)
         echo '<div class="cruxembed">';
         require_once(INCLUDES_PATH . '/include/CrUX.php');
 
-        InsertCruxHTML($testRunResults);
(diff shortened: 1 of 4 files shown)


www/vitals.php
```diff
@@ -151,13 +151,14 @@ function scrollTo(node) {
...
+
     global $testRunResults;
+
     if (isset($testRunResults)) {
         echo '<div class="cruxembed">';
         require_once(INCLUDES_PATH . '/include/CrUX.php');
 
-        InsertCruxHTML($testRunResults, null, "cwv");
-
+        InsertCruxHTML($testRunResults);
```
stoyan, line 161: this is the... crux of the change, the rest is cleanup

stoyan:
kept the output as-is, it makes sense to compare only test-measurable vitals with the ones from the wild

PR #2891, opened 2023-05-05
fix(getLocations): hide paid locations from unpaid users

jefflembeck:


stoyan (approved):
looks good!

Not sure about the unminification of co2.js, maybe leave as is with prettier ignore of some sort?

stoyan (approved):

PR #2892, opened 2023-05-06
Very small text changes to carbon control

fershad:
This PR makes a couple of minor text changes to the Carbon Control page.
It also adds the `&cc` param to the hyperlink which allows people to test again from another location.


scottjehl:
This is great, thanks Fershad!

scottjehl:
One minor thing that might be better: if we can change the url to use the /carbon-control slug instead, that's probably better.

I can make that edit.
`<a href="/carbon-control/?url={{ $test_url }}"`

PR #2911, opened 2023-05-24
Admin URL to dump test settings for debugging

stoyan:
e.g. /admin.php?page=test-info&test=221103_BiDcK9_7QN

### DIFF
--- a/www/src/Handlers/Admin.php
+++ b/www/src/Handlers/Admin.php
@@ -83,6 +83,14 @@ class Admin
 
     public static function getTestInfo(RequestContext $request_context): Response
     {
+
+        // psalm workaround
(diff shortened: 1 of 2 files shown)


www/src/Handlers/Admin.php
```diff
@@ -83,6 +83,14 @@ public static function cacheCheck(RequestContext $request_context): Response
 
     public static function getTestInfo(RequestContext $request_context): Response
     {
+
+        // psalm workaround
+        if (!function_exists('GetTestInfo')) {
```
jefflembeck, line 88: oh man, this is amazing. good for keeping stock of all the stuff we need to move, too

PR #2919, opened 2023-06-01
Revert "fix(account): logout bad clients"

jefflembeck:
This reverts commit e45253efb0b42abe541ed78c5f9609df4066b814.


jefflembeck:
This is causing issues. I'm dropping it.

PR #2950, opened 2023-07-19
fix: BUG 168255: removed fixed width from form inputs

alectejada:
Small fix to avoid showing scrollbars when using the browser zoom

before
before

before with 125% zoom
before_125_zoom

after
after

after 125% zoom
after_125_zoom


jefflembeck:
awesome. thanks!

jefflembeck (approved):

claud-io (approved):

PR #2952, opened 2023-07-19
fix: BUG 131913 stop losing the filter text when filtering the test history  

alectejada:
this change will prevent losing the text the user has on the text input when filtering the test history, it does so by making sure we keep it on the url query


www/testlog.php
```diff
@@ -52,7 +55,7 @@
         'local' => isset($_REQUEST['local']) && $_REQUEST['local'],
         'body_class' => 'history',
         'page_title' => 'WebPageTest - Test History',
-
+        'filter' => $filterstr
```
jefflembeck, line 58: I know you have the regex above, but I'd just use `htmlspecialchars()` https://www.php.net/htmlspecialchars

alectejada: hey jeff since its being rendered on a blade template it should be going through this function already, should we add anyway?

image

https://laravel.com/docs/10.x/blade

jefflembeck: oh that's right! nice catch. spaced the whole "this is on blade" thing.

jefflembeck (approved):

PR #2953, opened 2023-07-20
AZU174879: Update Messaging on Page Displayed when Anonymous User Hits Limit

claud-io:


www/templates/errors/runtest.php
```diff
@@ -1,6 +1,12 @@
...
-  <h1>Oops! <em>There was a problem with the test.</em></h1>
+  <h1>Oops! <em>
   <?php
-
-  echo $error; ?>
+    $title = "There was a problem with the test.";
+    if (strlen($errorTitle)) {
+	$title = $errorTitle ;
+    }
+    echo $title;
```
jefflembeck, line 8:
```suggestion
    echo $errorTitle ?? "There was a problem with the test.";
```

jefflembeck: in fact, you can make this whole thing:

```php
<?= $errorTitle ?? "There was a problem with the test." ?>
```

jefflembeck: Then, you can inline it with the things around it:

```php
<?= <h1>Oops! <em><?= $errorTitle ?? "There was a problem with the test." ?></em></h1> ?>
<?= $error ?>
```

PR #2958, opened 2023-08-09
removed text/tips and added mktg div

claud-io:


www/running.inc
```diff
@@ -11,6 +11,8 @@
             $noanalytics = true;
             $page_keywords = array('WebPageTest','Website Speed Test','Page Speed');
             $page_description = "Running website performance test.";
+	    
+	    require_once WWW_PATH . '/templates/layouts/google-tag-manager.inc';
```
claud-io, line 15: @mjkozicki  this is already included in head.inc but changing $noanalytics = false; will import also other scripts, so probably better to just include gtm here?

www/running.inc
```diff
@@ -11,7 +11,9 @@
             $noanalytics = true;
             $page_keywords = array('WebPageTest','Website Speed Test','Page Speed');
             $page_description = "Running website performance test.";
+	    
             include('head.inc');
+	    require_once WWW_PATH . '/templates/layouts/google-tag-manager.inc';
```
claud-io, line 16: @mjkozicki there is some css loaded after the gtm that is conflicting with it, moving after it fix the issue

mjkozicki (approved):

PR #2983, opened 2023-10-17
Fixed mispelled variables

lbartoli79:


claud-io (approved):
let's find a sample of before/after

PR #2993, opened 2023-10-30
Contact Information text is overlapping with Start Test button

v-hclaveria:
Proposed solution, pending to be tested on the QA environment.
Change on styling, adding missing tag. Applying the same change to password modal.
@claud-io 
@mjkozicki

### DIFF
--- a/www/assets/css/account.css
+++ b/www/assets/css/account.css
@@ -232,7 +232,7 @@ body.theme-b {
 .modal-open {
   visibility: visible;
   display: block;
-  z-index: 1000001;
+  z-index: 1001;
   background: rgb(244, 244, 244);
   box-sizing: border-box;
   padding: 5vh;
--- a/www/assets/css/account.css
+++ b/www/assets/css/account.css
@@ -189,6 +189,20 @@ body.theme-b {
   border: 0;
 }
 
+.contact-info-modal {
+    height: 38rem !important;
+}
+.contact-info-modal-form {
+    height: 26rem !important;
+}
+
+.password-modal {
+    height: 34rem !important;
+}
+.password-modal-form {
+    height: 22rem !important;
+}
+
 .modal,
 .modal_screen {
   position: fixed;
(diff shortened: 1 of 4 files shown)


v-hclaveria:
image

www/templates/account/includes/modals/contact-info.php
```diff
@@ -31,9 +31,11 @@
             <input type="hidden" name="id" value="<?= $id; ?>" />
             <input type="hidden" name="type" value="contact-info" />
             <input type="hidden" name="csrf_token" value="<?= $csrf_token ?>" />
-            <div class="save-button">
-                <button type="submit" class="pill-button blue">Save</button>
-            </div>
+            <div class="section">
+				<div class="save-button">
```
claud-io, line 35: please format

www/assets/css/account.css
```diff
@@ -189,6 +189,20 @@ body.theme-b {
   border: 0;
 }
 
+.contact-info-modal {
+    height: 38rem !important;
```
claud-io, line 193: please check with different resolutions and format code

www/assets/css/account.css
```diff
@@ -218,7 +232,7 @@ body.theme-b {
 .modal-open {
   visibility: visible;
   display: block;
-  z-index: 1001;
+  z-index: 1000001;
```
lbartoli79, line 235: Look like this is not correct, this value should be greater than **wpt-header** z-index too, but it fails when we have different media width:
- default z-index: https://github.com/catchpoint/WebPageTest/blob/cd6502f48cac364f3b280fbdd7b84579a1538ab7/www/assets/css/wpt-header.css#L38
- z-index with `@media (min-width: 86em)`: https://github.com/catchpoint/WebPageTest/blob/cd6502f48cac364f3b280fbdd7b84579a1538ab7/www/assets/css/wpt-header.css#L419

www/assets/css/wpt-header.css
```diff
@@ -416,7 +416,7 @@ wpt-header .wptheader_nav_menu details[open] summary span:after {
 
 @media (min-width: 86em) {
   wpt-header {
-    z-index: 999999;
+    z-index: 1000;
```
lbartoli79, line 419: This is going to change header z-index everywhere, are you sure it does not break anything?

PR #2994, opened 2023-10-31
Overlapping text (Re-open #2993)

v-hclaveria:
Re-opening #2993. Apparently, it closes at re-naming the branch

With the first commit, I intended not to break anything and test how it behaves and look for impacts, since there are places I cannot check locally. It truly seems rare this new change won't cause any break, but those places are what I'm trying to find.
Please let me know if there's a more certain way to solve this that I'm not seeing.

@claud-io

### DIFF
--- a/www/assets/css/account.css
+++ b/www/assets/css/account.css
@@ -232,7 +232,7 @@ body.theme-b {
 .modal-open {
   visibility: visible;
   display: block;
-  z-index: 1000001;
+  z-index: 1001;
   background: rgb(244, 244, 244);
   box-sizing: border-box;
   padding: 5vh;
--- a/www/assets/css/account.css
+++ b/www/assets/css/account.css
@@ -189,6 +189,20 @@ body.theme-b {
   border: 0;
 }
 
+.contact-info-modal {
+    height: 38rem !important;
+}
+.contact-info-modal-form {
+    height: 26rem !important;
+}
+
+.password-modal {
+    height: 34rem !important;
+}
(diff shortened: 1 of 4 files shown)


www/assets/css/wpt-header.css
```diff
@@ -416,7 +416,7 @@ wpt-header .wptheader_nav_menu details[open] summary span:after {
 
 @media (min-width: 86em) {
   wpt-header {
-    z-index: 999999;
+    z-index: 1000;
```
claud-io, line 419: those changes are going to impact header everywhere have you checked with multiple resolutions?

www/assets/css/account.css
```diff
@@ -189,6 +189,20 @@ body.theme-b {
   border: 0;
 }
 
+.contact-info-modal {
+    height: 38rem !important;
```
claud-io, line 193: is important mandatory?

www/templates/account/includes/modals/contact-info.php
```diff
@@ -31,8 +31,10 @@
             <input type="hidden" name="id" value="<?= $id; ?>" />
             <input type="hidden" name="type" value="contact-info" />
             <input type="hidden" name="csrf_token" value="<?= $csrf_token ?>" />
-            <div class="save-button">
-                <button type="submit" class="pill-button blue">Save</button>
+            <div class="section">
```
claud-io, line 34: what's the effect of this section class?

PR #2998, opened 2023-11-06
Bugfix: added interval to set correct select value after form autocomplete

alectejada:
this change should fix the problem on the country select from the user data form(step-2), where the browser autocomplete is setting the value the user selected previously but is not triggering an event when doing so.

this is a fix where we would check the values on an interval and when a change is detected we manually set the correct states.


alectejada:
closing this, we decided to just disable autocomplete on the fields.

PR #3004, opened 2023-11-09
Updated iPhone/iPad mobile devices

lbartoli79:


pmeenan:
Did you mean to change the trhottle factor for the existing Android devices? Those are the baselines that you should be setting other throttles from because they were calibrated on real devices.

lbartoli79:
Those two files www/settings/mobile_devices.ini docker/local/server/config/mobile_devices.ini are different, which one is correct?

pmeenan:
I'm not sure why the docker container has a separate copy but the `www/settings/mobile_devices.ini` one is the authoritative file (and the one used by production).

Change LGTM

pmeenan (approved):

lbartoli79:
@pmeenan Thanks for the clarification. I'm going to open another PR to align docker configuration with the one used in prod

claud-io (approved):

PR #3006, opened 2023-11-10
Add message for lighthouse

claud-io:


www/resources/views/pages/lighthouse.blade.php
```diff
@@ -73,6 +73,7 @@ function copySelector(selector) {
                     <span class="lh_score_grade-average">50-89</span>
                     <span class="lh_score_grade-pass">90-100</span>).
                 </p>
+                <p>For scripted tests, Lighthouse report will only be available on the final step.</p>
```
lbartoli79, line 76: Did you forget something?

> Custom Script tab should display following message:
> Message: "Lighthouse does not perform per step reports for custom scripting, but will run a report on the final step."

lbartoli79 (approved):

PR #3008, opened 2023-11-10
Fix location when there is only a single browser

lbartoli79:
Fix invalid location error when running chrome emulated device test on location having only one browser defined.


claud-io (approved):

pmeenan (approved):
LGTM, thanks

PR #3010, opened 2023-11-10
Bugfix: added js to sync country and state fields

alectejada:
Added javascript to have the country selector changes interact with the state select. its the same logic we have on the signup pages.

### DIFF
--- a/www/templates/account/billing/update-payment-confirm-address.php
+++ b/www/templates/account/billing/update-payment-confirm-address.php
@@ -78,7 +78,7 @@
         <input type="hidden" name="type" value="update-payment-method-confirm-billing" />
         <button class="yellow pill-button" type="submit">Confirm Billing Address</button>
     </form>
-    <script src="/assets/js/country-list/country-list.js"></script>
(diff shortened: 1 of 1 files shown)


www/templates/account/billing/update-payment-confirm-address.php
```diff
@@ -78,4 +78,25 @@
         <input type="hidden" name="type" value="update-payment-method-confirm-billing" />
         <button class="yellow pill-button" type="submit">Confirm Billing Address</button>
     </form>
+    <script defer src="/assets/js/country-list/country-list.js?v=<?= constant('VER_JS_COUNTRY_LIST') ?>"></script>
```
claud-io, line 81: @mjkozicki this is same script we have in signup page

claud-io: @alectejada please copy the exact scirpt, why you need the php  tag?

alectejada: this is the script that is on the billing-address-form.php is the one we use on signup.

claud-io (approved):

PR #3021, opened 2023-12-06
Pixel device update

lbartoli79:
Updated mobile_devices.ini

### DIFF
--- a/www/settings/mobile_devices.ini
+++ b/www/settings/mobile_devices.ini
@@ -73,6 +73,107 @@ dpr=3.5
 throttle_cpu=4
 ua="Mozilla/5.0 (Linux; Android 8.1.0; Pixel 2 XL Build/OPM1.171019.011) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.54 Mobile Safari/537.36"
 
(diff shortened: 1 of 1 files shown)


lbartoli79:
@pmeenan for those we set `throttle_cpu=1` , do you think we are ok with this value?

pmeenan:
> @pmeenan for those we set `throttle_cpu=1` , do you think we are ok with this value?

For the Pixel 6 and later, yes but for the earlier devices, no. The pixel 2 has a "4" throttle factor and looking at the geekbench single-threaded performance for the [Pixel 3](https://browser.geekbench.com/android_devices/google-pixel-3), [Pixel 4](https://browser.geekbench.com/android_devices/google-pixel-4) and [Pixel 5](https://browser.geekbench.com/android_devices/google-pixel-5), they are all less than 4x faster than the [Pixel 2](https://browser.geekbench.com/android_devices/google-pixel-2).

If you don't have access to actual devices to benchmark, my recommendation would be to scale the Pixel 2 throttle factor by how fast each of the other devices is relatively in geekbench.

pmeenan (approved):
LGTM

PR #3022, opened 2023-12-08
Bugfix: added private form value to test payload

alectejada:
small fix, added private value to test payload. this should fix the problem where the private setting is not set on the .ini file


claud-io:
Going to merge so we can try it

claud-io (approved):

PR #3026, opened 2023-12-12
Bugfix: contact info is readonly for verified user

alectejada:
This should make the inputs on the contact info modals read only and hide the save button for non verified users.

image


claud-io (approved):
LGTM

PR #3029, opened 2023-12-15
Adjustment z-index account css

v-hclaveria:


www/assets/css/account.css
```diff
@@ -293,6 +293,7 @@ body.theme-b {
     left: 50%;
     transform: translate(-50%, -50%);
     max-width: 800px;
+    z-index: 10000001;
```
claud-io, line 296: only for media query?

v-hclaveria: Yes, it seems only necessary for wide screens

claud-io: image

claud-io (approved):

PR #3032, opened 2023-12-22
Update vendors

claud-io:


lbartoli79 (approved):

pmeenan:
Which libraries were updated? Did you test that the pages they are used on still work ok?  With 2781 files changed It's not something anyone is going to be able to review manually so being clear about which libraries were upgraded in the commit message will be critical.

claud-io:
> Which libraries were updated? Did you test that the pages they are used on still work ok? With 2781 files changed It's not something anyone is going to be able to review manually so being clear about which libraries were upgraded in the commit message will be critical.

@pmeenan that's an update for all vendors related to this change https://github.com/catchpoint/WebPageTest/pull/3016.
We are still evaluating this change, for sure this is going to require a good regression for all the pages. (we checked main pages and was not able to see anything breaking but need a better check on everything else)

PR #3050, opened 2024-03-14
Embed the perfetto trace UI directly

pmeenan:
This embeds the Perfetto UI for viewing traces directly using the [supported embedding mechanism](https://perfetto.dev/docs/visualization/deep-linking-to-perfetto-ui) instead of maintaining a copy of the code that needs to be kept up to date.


pmeenan:
Hmm, looks like this picked up a few extra committs, let me update and cherry-pick

PR #3051, opened 2024-03-14
Embed the Perfetto UI directly

pmeenan:
This embeds the perfetto UI for viewing Chrome traces directly into WebPageTest using the [supported embedding mechanism](https://perfetto.dev/docs/visualization/deep-linking-to-perfetto-ui) instead of maintaining a copy of the code that needs to be updated manually.


www/chrome/trace.php
```diff
@@ -6,33 +6,42 @@
...
+            }}, ORIGIN);
     }
-    readfile($conclusion);
-} else {
-    header("HTTP/1.0 404 Not Found");
-}
+  </script>
+</head>
+<body style="margin:0px;padding:0px;overflow:hidden">
+  <?php
```
mjkozicki, line 43: Any reason to have this as `echo`?

pmeenan: oh, nope - I'll update it. It's just left over from copying the dev tools embed.

mjkozicki (approved):

Issue #3042, opened 2024-02-09
Add ability to specify min/start time for custom waterfalls


jspurlin:
## Feature Request Summary
This is a request to add the ability to set a min/start time for custom waterfalls. Today you can specify a max/end time, but no start time.

### What's the motivation or use case for the change/feature?
I want the ability to "zoom in" on a specific chunk of time from the waterfall, say 2-3.5s for example. Today I can only see 0-3.5s

slightlyoff:
This is a request coming from MSFT's use of WPT internally. I don't have edit permissions on the bug, but if the owners here feel like it would be best for us to contribute patches in support of this, let us know.

/cc @pmeenan @andydavies

pmeenan:
FWIW, I'm happy to review PR's or answer any questions (or finally take a crack at it myself) but I don't have repository permissions to merge so that part would depend on Catchpoint.

PR #3057, opened 2024-04-11
CustomWaterfall: add min functionality

jspurlin:
Prior to this PR users could only set a max time to filter on when creating custom waterfalls. This meant that custom waterfalls always started at t=0 even if what you cared about didn't start until the t=5s mark

This PR adds the ability to optionally specify a min time to filter on in addition to setting a max and requests to filter on.

I verified that the custom filter scales correctly in app parts (Requests, CUP Utilization, Bandwidth, Browser Main Thread, and Long Tasks)

For example, if I care about the 5-10s range of the waterfall this is what it looks like

# Before
image

# After
image

This is a fix for [Add ability to specify min/start time for custom waterfalls #3042](https://github.com/catchpoint/WebPageTest/issues/3042)

### DIFF
--- a/www/waterfall.inc
+++ b/www/waterfall.inc
@@ -326,7 +326,11 @@ function _GetMapOrImage($rows, $url, $page_events, $options, $is_image_map, $pag
     $data_width = $width - $data_x - 3;
 
     // Figure out the scale.
+    $min_ms = 0;
     $max_ms = 0;
+    if (@$_REQUEST['min'] > 0) {
+        $min_ms = (int)($_REQUEST['min'] * 1000.0);
+    }
     if (@$_REQUEST['max'] > 0) {
         $max_ms = (int)($_REQUEST['max'] * 1000.0);
     } else {
@@ -342,7 +346,7 @@ function _GetMapOrImage($rows, $url, $page_events, $options, $is_image_map, $pag
             }
         }
     }
-    $x_scaler = new XScaler($max_ms, $data_x + 1, $data_width);
+    $x_scaler = new XScaler($min_ms, $max_ms, $data_x, $data_width);
 
     // get any user timings
     if (!isset($options['show_user_timing']) || $options['show_user_timing']) {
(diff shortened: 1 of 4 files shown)


jspurlin:
@pmeenan / @vibaldem  FYI, this PR is for #3042

arthesh:
Tagging @mjkozicki , @claud-io , would appreciate if you could review this PR and provide feedback.

pmeenan (approved):
LGTM.

`Minimum` time feels a bit off and `Start` seems more appropriate but given that the end is called `Maximum` it is appropriate to keep it aligned.

Thanks for doing this, it's been a long-standing request and I've been dreading going through the scaling code. The change looks nice and clean.

jspurlin:
@mjkozicki ...PING... Can you review (and merge if things look good)?

jspurlin:
@mjkozicki @claud-io  ...PING, PING... Can you review (and merge if things look good)?

claud-io:
Hello @jspurlin we are testing it, will let you know once is done. Thanks

claud-io (changes requested):
@jspurlin seems page sometimes break when min >= max or with neg numbers, can you please add check for those cases? thanks

claud-io:
@jspurlin thanks, going to start testing it

www/waterfall.inc
```diff
@@ -326,7 +326,11 @@ function _GetMapOrImage($rows, $url, $page_events, $options, $is_image_map, $pag
     $data_width = $width - $data_x - 3;
 
     // Figure out the scale.
+    $min_ms = 0;
     $max_ms = 0;
+    if (@$_REQUEST['min'] > 0 && @$_REQUEST['min'] < @$_REQUEST['max']) {
```
lbartoli79, line 331: Can you address `$max_ms` effective value? As is `min` it only works if a `max` value is provided

lbartoli79 (changes requested):
We should not draw vertical lines on first column

image

jspurlin:
> We should not draw vertical lines on first column
> 
> image

@lbartoli79 Done. I've made sure to account for the case where min is provided but max is not and not to draw lines < min_ms

jspurlin:
...PING...

@claud-io and @lbartoli79

lbartoli79 (approved):

PR #3084, opened 2024-06-28
Update insecure-requests.inc

Siddhantshukla814:
if URL is a certificate URL then it should not be considered as an insecure request.

### DIFF
--- a/www/experiments/insecure-requests.inc
+++ b/www/experiments/insecure-requests.inc
@@ -1,5 +1,27 @@
 <?php
 
+function isCertificateUrl($url, $contentType) {
+    $certificateContentTypes = [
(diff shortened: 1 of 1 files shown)


www/experiments/insecure-requests.inc
```diff
@@ -1,14 +1,38 @@
 <?php
 
+function isCertificateUrl($url, $contentType) {
```
lbartoli79, line 3: Remove $url, it's not needed anymore

Siddhantshukla814: Pardon I left it.

lbartoli79 (approved):

PR #3149, opened 2025-02-27
Issue 3148: Change expire response header to 7 days later

imkalvin:


www/common_lib.inc
```diff
@@ -2639,7 +2639,7 @@ function json_response(&$response, $allow_crossorigin = true)
 {
     header("Content-type: application/json; charset=utf-8");
     header("Cache-Control: no-cache, must-revalidate", true);
-    header("Expires: Sat, 26 Jul 1997 05:00:00 GMT");
+    header("Expires: " . gmdate("D, d M Y H:i:s", time() + 7 * 24 * 60 * 60) . " GMT");
```
lbartoli79, line 2642: This is not only used for json results, not sure the expires value here is good for all the use cases involved

claud-io: @imkalvin bump on this

claud-io:
@imkalvin closing since waiting for a reply since march. Please fill free to re-open it once you are able to address the comment

PR #3165, opened 2025-09-21
Update Dockerfile-wptagent with git target folder

karoltheguy:
Changed target folder for git clone as it was renamed in previous change. 

**From** https://github.com/WPO-Foundation/wptagent.git **to** https://github.com/catchpoint/WebPageTest.agent.git
This prevented the dockerfile to compile as the root folder to became WebPageTest.agent instead of wptagent.

Root folder is now back to wptagent


colorful-tones:
Just hit this issue, and confirming this PR fixes it.

